PR #5, opened 2020-05-07
Added support for multiple database connections

### LINKED ISSUE #4: DB transactions with multiple connections
Strategies that utilize database transactions do not work correctly with multiple database connections configured. It always uses the default connection:

[DatabaseTransactionHelpers.php](https://github.com/knuckleswtf/scribe/blob/master/src/Extracting/DatabaseTransactionHelpers.php)
```
app('db')->beginTransaction();
```

I started adding support for multiple connections in [my fork](https://github.com/hendrikbl/scribe/tree/multi-connection). It simply gets the connection from the model type and passes it to the `startDbTransaction` and `endDbTransaction` function.

I'm not sure though how to do it in [ResponseCalls.php](https://github.com/knuckleswtf/scribe/blob/master/src/Extracting/Strategies/Responses/ResponseCalls.php) since I don't know any way to get the utilized connection from the route itself. One way could be simply iterating over all connections and starting and ending transactions for all connections. But if there is a better method I'd prefer that :D

Let me know what you think and if i missed anything.

hendrikbl:
Closes #4 

For ApiResource and Transformer strategies the model types connection is used to start the transaction. For the ResponseCall strategy a transaction for every database connection is started since we don't know which one is used for a specific route.

### DIFF
--- a/src/Extracting/DatabaseTransactionHelpers.php
+++ b/src/Extracting/DatabaseTransactionHelpers.php
@@ -7,23 +7,27 @@ use Exception;
 trait DatabaseTransactionHelpers
 {
     /**
+     * @param string $connection
+     *
      * @return void
      */
-    private function startDbTransaction()
+    private function startDbTransaction(String $connection = null)
     {
         try {
-            app('db')->beginTransaction();
+            app('db')->connection($connection)->beginTransaction();
         } catch (Exception $e) {
         }
     }
 
     /**
+     * @param string $connection
+     *
      * @return void
      */
-    private function endDbTransaction()
+    private function endDbTransaction(String $connection = null)
     {
         try {
-            app('db')->rollBack();
+            app('db')->connection($connection)->rollBack();
         } catch (Exception $e) {
         }
     }
--- a/src/Extracting/DatabaseTransactionHelpers.php
+++ b/src/Extracting/DatabaseTransactionHelpers.php
@@ -11,11 +11,15 @@ trait DatabaseTransactionHelpers
      *
      * @return void
      */
-    private function startDbTransaction(String $connection = null)
+    private function startDbTransaction()
     {
-        try {
-            app('db')->connection($connection)->beginTransaction();
-        } catch (Exception $e) {
+        $connections = array_keys(config('database.connections', []));
+
+        foreach ($connections as $conn) {
+            try {
+                app('db')->connection($conn)->beginTransaction();
+            } catch (Exception $e) {
+            }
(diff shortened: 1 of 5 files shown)


src/Extracting/DatabaseTransactionHelpers.php
```diff
@@ -7,23 +7,27 @@
 trait DatabaseTransactionHelpers
 {
     /**
+     * @param string $connection
+     *
      * @return void
      */
-    private function startDbTransaction()
+    private function startDbTransaction(String $connection = null)
```
shalvah, line 14: Canonically, in PHP, it's `string`, not `String`. Let's keep it uniform.

src/Extracting/Strategies/Responses/ResponseCalls.php
```diff
@@ -90,7 +90,11 @@ public function makeResponseCall(Route $route, array $context, array $rulesToApp
      */
     private function configureEnvironment(array $rulesToApply)
     {
-        $this->startDbTransaction();
+        // Start transactions for all connections since we don't know wich one is used
+        $connections = array_keys(config('database.connections'));
```
shalvah, line 94: Can we rather move this whole logic into the startDbTransactions method? Same for the endDbTransactions.The configureEnvironment method should just be a high-level list of steps.

Also, don't forget to wrap each individual call to app('db')->... in a try/catch.

shalvah:
Another issue iterating over connections would solve: models that don't extend Eloquent.

src/Extracting/DatabaseTransactionHelpers.php
hendrikbl: If we take the other approach for everything, we don't need this anyways :)

hendrikbl:
Let me know if there is something left to do :)

src/Extracting/Strategies/Responses/ResponseCalls.php
hendrikbl: Done!

tests/Fixtures/TestModel.php
```diff
@@ -2,12 +2,14 @@
...
 
+use Illuminate\Database\Eloquent\Model;
+
 /**
  * A demo test model.
  *
  * @author Tobias van Beek <t.vanbeek@tjvb.nl>
  */
-class TestModel
+class TestModel extends Model
```
shalvah, line 12:
```suggestion
class TestModel
```

PR #9, opened 2020-05-13
Update install instructions to use composer require

ncatanchin:


shalvah:
Oh damn, too much JavaScript. 😄

shalvah:
It should even be `require --dev`.

PR #10, opened 2020-05-13
Update PostmanCollectionWriter.php

patrickcurl:
in my config for headers I have @{{token}} which in the views shows up as {{token}}, but I want to use {{token}} in collection for postman w/out having to search/replace. So the solution was this.

### DIFF
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -106,6 +106,8 @@ class PostmanCollectionWriter
                 'Accept' => 'application/json',
             ])
             ->map(function ($value, $header) {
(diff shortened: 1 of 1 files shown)


retry19:
@shalvah how to add like this for **body param** in controller file? it's possible? 

For example :
- on PHP
    ```php
    /**
    * @bodyParam username string required Example: @{{username}}
    * /
    ```
- on Postman
    ```json
    "username": "{{username}}"
    ```

But, for now if we add `@{{username}}` in controller, in postman will also be the same as that, unchanged

And if we add `{{username}}` in controller (without `@`), in view will be error, because it's will define as constant

PR #13, opened 2020-05-18
form-data support

ErnestStaug:
This will generate proper json for  postman if request is form-data otherwise will default to raw json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,8 @@
     },
     "config": {
         "preferred-install": "dist",
-        "sort-packages": true
+        "sort-packages": true,
+        "process-timeout": 600
     },
     "replace": {
         "mpociot/laravel-apidoc-generator": "*"
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -90,13 +90,13 @@ class PostmanCollectionWriter
     {
 
         $body = [];
-        $contentType = $route['headers']['Content-Type'];
+        $contentType = $route['headers']['Content-Type'] ?? null;
         switch ($contentType) {
             case 'multipart/form-data':
                 $mode = 'formdata';
                 break;
             case 'application/json':
-            case 'default':
+            default:
                 $mode = 'raw';
         }
         $body['mode'] = $mode;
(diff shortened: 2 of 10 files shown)


shalvah:
Nice, thank you. I completely forgot about form data support in the Postman collection.

shalvah:
You need to fix the build though.

ErnestStaug:
> You need to fix the build though.

yes, sorry forgot about tests, quite new to open source :)

shalvah:
That's nice.👏👏 Did you read the contribution guide? https://scribe.readthedocs.io/en/latest/contributing.html

ErnestStaug:
> That's nice.👏👏 Did you read the contribution guide? https://scribe.readthedocs.io/en/latest/contributing.html

Now I have :D, I've put some unit tests and fixed issue. should be all good now ;)

composer.json
```diff
@@ -67,7 +67,8 @@
     },
     "config": {
         "preferred-install": "dist",
-        "sort-packages": true
+        "sort-packages": true,
+        "process-timeout": 600
```
shalvah, line 71: Why this?

Tip: when you make changes like this, that are likely not going to be immediately apparent to others, please add an explanation in the description of your PR.

src/Writing/PostmanCollectionWriter.php
```diff
@@ -69,28 +69,66 @@ public function makePostmanCollection()
...
-                    $mode => json_encode($route['cleanBodyParameters'], JSON_PRETTY_PRINT),
-                ],
+                'body' => $this->getBodyData($route),
                 'description' => $route['metadata']['description'] ?? null,
                 'response' => [],
             ],
         ];
     }
 
+    protected function getBodyData($route): array
```
shalvah, line 89: For completeness:

```suggestion
    protected function getBodyData(array $route): array
```

composer.json
ErnestStaug: increased timeout because my machine was getting composer timeout errors when running tests... I own macbook pro 2016, it's not the slowest computer but certain tests were timing out. ( default timeout is 300, so I don't think this needs much explanation)

shalvah: Fair.

> default timeout is 300, so I don't think this needs much explanation

Sometimes the most innocuous changes can have significant effects.🙂

PR #23, opened 2020-05-23
Resolves #22 - Set the correct config key for documentation URL

### LINKED ISSUE #22: Config key mismatch in laravel routes
The route configuration is looking for config key `scribe.laravel.docs_url` 

https://github.com/knuckleswtf/scribe/blob/d8a1e1713bbf10295063d47981566f2df746d5bb/routes/laravel.php#L5

However, the default configuration file defines it as `scribe.laravel.url`

https://github.com/knuckleswtf/scribe/blob/d8a1e1713bbf10295063d47981566f2df746d5bb/config/scribe.php#L36

I did not submit a PR as unsure of which option you would prefer to go ahead with!

---

Thanks for all the hard work on the new package! :) After some initial hurdles, I just got my docs building this morning and ❤️ the design improvements. Jumping in to take advantage of the feature enhancements now. 🥳

ncatanchin:
As per #22 - updated base configuration file

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -33,7 +33,7 @@ return [
          * URL path to use for the docs endpoint (if `add_routes` is true).
          * By default, `/docs` opens the HTML page, and `/docs.json` downloads the Postman collection.
          */
-        'url' => '/docs',
+        'docs_url' => '/docs',
 
         /*
          * Middleware to attach to the docs endpoint (if `add_routes` is true).


shalvah:
How about the documentation too?

ncatanchin:
Will do!

ncatanchin:
Unless I'm missing something, docs appear to already be correct

image

Let me know if anything else needed, cheers

shalvah:
Cool then. 👍

PR #31, opened 2020-05-28
feat(config): ability to set postman base_url independently

marandy0905:
In the [document](https://github.com/knuckleswtf/scribe/blob/6863189588e4a9a1269c5c9b41de6a9905d80131/docs/generating-documentation.md#L26) mentioned that a `postman.base_url` can be specified for postman collection generated.
However after reviewing the code, I find only the `base_url` in the outer array can be changed and by changing it will change both the postman collection's `base_url` and the examples.

Since I think many users might preferred having different customization of page and postman, I've tried to implement this feature.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -123,6 +123,13 @@ INTRO
         'enabled' => true,
 
         /*
+         * The base URL to be used in the Postman collection.
+         * This will override the base_url settings above only for Postman.
+         * If this is null, Scribe will use the value of config('app.url').
+         */
+        'base_url' => null,
+
+        /*
          * The description for the exported Postman collection.
          */
         'description' => null,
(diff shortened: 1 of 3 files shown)


shalvah:
Fair, but what's a use case where they'd want a separate Postman base URL? Using variables?

marandy0905:
@shalvah definitely! I think lot's of QAs use postman a lot, and by having the ability of setting variables will be a great option for them to setup same test case in different environments.

marandy0905:
An use case that I use a lot is specifying my postman routes as something like:
`{{url}}/api/v1/XXX`, therefore the `{{url}}` can be switched each time when I'm in beta or alpha phase by not regenerating almost identical collections.

shalvah:
👍

shalvah:
In the future, you should also [update the docs](https://github.com/knuckleswtf/scribe/commit/052879dd1af38960a56e0239ad6f2f88ebb01bc0).

marandy0905:
thanks! missed that part!

PR #43, opened 2020-06-18
Resolves #40 - Include raw in requests.url

### LINKED ISSUE #40: Include raw in request.url?
Hi,

I'm using Insomnia, and when I import _collection.json_ the urls to the endpoints are empty.

Importing to Postman, exporting from there and importing again to Insomnia makes be believe that this could be fixed by adding _raw_ to the export:
```
"item": [
{
  "name": "/myendpoint",
  "request": {
    "url": {
      **"raw": "https://example.com/myendpoint",**
      "protocol": "https",
      "host": "example.com",
      "path": "/myendpoint",
      "query": []
    },
}
]
```

Should this be added, to conform to the way Postman exports its endpoints?

chriserikhaugli:
Fixes import to Insomnia

Fixes #40 

> **Please read the [contribution guidelines](https://scribe.readthedocs.io/en/latest/contributing.html), especially the section on making a pull request, before creating a PR.** Otherwise, your PR may be turned down.

### DIFF
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -183,6 +183,14 @@ class PostmanCollectionWriter
             })->values()->toArray(),
         ];
 
+        // Create raw url-parameter (Insomnia uses this on import)
+        $query = collect($base['query'] ?? [])->map(function ($array) {
+            return $array['key'].'='.$array['value'];
+        })->implode('&');
+        $base['raw'] = sprintf('%s://%s/%s%s',
+            $base['protocol'], $base['host'], $base['path'], $query ? '?'.$query : null
+        );
+
         // If there aren't any url parameters described then return what we've got
         /** @var $urlParams Collection */
         if ($urlParams->isEmpty()) {
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -184,8 +184,8 @@ class PostmanCollectionWriter
         ];
 
         // Create raw url-parameter (Insomnia uses this on import)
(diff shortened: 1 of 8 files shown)


src/Writing/PostmanCollectionWriter.php
```diff
@@ -183,6 +183,14 @@ protected function makeUrlData($route)
             })->values()->toArray(),
         ];
 
+        // Create raw url-parameter (Insomnia uses this on import)
+        $query = collect($base['query'] ?? [])->map(function ($array) {
```
shalvah, line 187: Let's be a little more semantic in our naming. Let\s rename this from `$array` to something more descriptive, like `$queryParamData`.

tests/Fixtures/collection.json
```diff
@@ -59,7 +60,8 @@
                             "protocol": "http",
                             "host": "localhost",
                             "path": "api\/withResponseTag",
```
shalvah, line 62: It probably works the same, but I'm wondering why the `path` has its slash escaped (probably unnecessary, since it's a forward slash), while your raw URL doesn't. Any idea why?

shalvah:
hey @chriserikhaugli got time to address these?

tests/Fixtures/collection.json
chriserikhaugli: No, I'm not sure.

But I see that the last commit on path says "_Adjusting existing tests for new format and fixing encoding issue_". 

If there might be encoding-issues it may be best to encode my raw url as well?

shalvah: No worries. We'll leave it as is for now.

PR #55, opened 2020-07-03
[WIP] [#48] Added defensive code for DatabaseConnection drivers that don't support transactions

### LINKED ISSUE #48: Mongodb Transaction Error
i m using mongodb in my project. So when ı generate `php artisan scribe:generate`, i get this error:
mongodb_transaction
How can i solve this error ? Is there any setting to disable mongodb connection? or to override `startDbTransaction()` that is trait method ?

Smudge3806:
Some third-party database drivers don't support transactions *cough* Jenssegers *cough*. I'm sure there are probably some more out there.

Thought I'd throw this out there as a possible solution, and get guidance from maintainers before writing the relevant Unit tests. 

Potentially fixes #48 
> **Please read the [contribution guidelines](https://scribe.readthedocs.io/en/latest/contributing.html), especially the section on making a pull request, before creating a PR.** Otherwise, your PR may be turned down.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -331,4 +331,9 @@ INTRO
      *
      */
     'routeMatcher' => \Knuckles\Scribe\Matching\RouteMatcher::class,
+
+    /**
+     * [Advanced usage] If a database driver does not support transactions, you can list it here to allow it to run.
+     */
+    'run_without_database_transactions' => [],
 ];
--- a/src/Extracting/DatabaseTransactionHelpers.php
+++ b/src/Extracting/DatabaseTransactionHelpers.php
@@ -3,6 +3,9 @@
 namespace Knuckles\Scribe\Extracting;
 
 use Exception;
+use Knuckles\Scribe\Exceptions\DbTransactionSupportException;
+use Knuckles\Scribe\Exceptions\ScribeException;
+use Knuckles\Scribe\Tools\ConsoleOutputUtils as c;
 
 trait DatabaseTransactionHelpers
 {
@@ -15,14 +18,22 @@ trait DatabaseTransactionHelpers
 
         foreach ($connections as $conn) {
             try {
-                tap(
-                    app('db')->connection($conn),
-                    function ($driver) {
-                        if (self::driverSupportsTransactions($driver)) {
-                            $driver->beginTransaction();
-                        }
-                    }
-                );
+                $driver = app('db')->connection($conn);
+
+                if (self::driverSupportsTransactions($driver)) {
+                    $driver->beginTransaction();
+
+                    return;
+                }
+
+                if ($this->isNoTransactionSupportAllowed($conn)) {
+                    throw DbTransactionSupportException::create($conn, get_class($driver));
+                }
+
+                c::warn("Database driver for the connection [{$conn}] does not support transactions!");
+                c::warn("Any changes made to your database will persist!");
+            } catch (ScribeException $e) {
+                throw $e;
             } catch (Exception $e) {
             }
         }
@@ -37,14 +48,16 @@ trait DatabaseTransactionHelpers
 
         foreach ($connections as $conn) {
             try {
-                tap(
-                    app('db')->connection($conn),
-                    function ($driver) {
-                        if (self::driverSupportsTransactions($driver)) {
-                            $driver->rollBack();
-                        }
-                    }
-                );
+                $driver = app('db')->connection($conn);
+
+                if (self::driverSupportsTransactions($driver)) {
+                    $driver->rollBack();
+
+                    return;
+                }
+
+                c::warn("Database driver for the connection [{$conn}] does not support transactions!");
+                c::warn("Any changes made to your database will persist!");
             } catch (Exception $e) {
             }
         }
@@ -69,4 +82,31 @@ trait DatabaseTransactionHelpers
 
         return true;
     }
+
+    /**
+     * Assesses whether drivers without transaction support can proceed
+     *
+     * @param string $connection_name Name of the connection
+     *
+     * @return boolean
+     */
+    private function isNoTransactionSupportAllowed(string $connection_name)
+    {
+        $config = $this->getConfig();
+
+        $allow_list = $config->get('run_without_database_transactions', false);
+
+        if (is_array($allow_list)) {
+            return in_array($connection_name, $allow_list);
+        }
+
+        return false;
+    }
+
+    /**
+     * Returns an instance of the documentation config
+     *
+     * @return DocumentationConfig
+     */
+    abstract public function getConfig();
 }
--- a/src/Extracting/DatabaseTransactionHelpers.php
+++ b/src/Extracting/DatabaseTransactionHelpers.php
@@ -15,7 +15,14 @@ trait DatabaseTransactionHelpers
 
         foreach ($connections as $conn) {
             try {
-                app('db')->connection($conn)->beginTransaction();
+                tap(
+                    app('db')->connection($conn),
+                    function ($driver) {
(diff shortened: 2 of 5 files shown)


shalvah:
That's a good fix. Of course, we'd need to emit a warning at the very least. But do we proceed? What happens if a dev is using a non-test db? And hope likely is that to happen? Or is it better to fail and have a CLI flag that allows use without transactions?

Smudge3806:
@shalvah Wrapped it in a layer of configuration that is hopefully extendable to future issues regarding databases.

I also added warnings to both the `beginDbTransaction` and `endDbTransaction` methods that to remind the user of the problems associated with allowing database drivers that don't support transactions.

If this meets with your general approval, I will update the documentation

config/scribe.php
```diff
@@ -331,4 +331,23 @@
...
+     * [Advanced usage] Customize how Scribe interacts with your databases.
+     */
+    'databases' => [
+        /**
+         * [Advanced usage] If you would like to enable the use of database drivers which don't support transactions
+         * set this value to true.
+         *
+         * NB: Changes made to your database during generatation WILL persist!
+         */
+        'allow_no_transactions' => false,
```
shalvah, line 345: I think `allow_no_transactions` is somewhat confusing. How about `use_without_transactions`?

Also, I'd rather we not nest this. We can also collapse these two configs into one. Something like:

```php
'run_without_database_transactions' => [
    ],
```

Then the user can add whatever connections they're okay with.

src/Extracting/Strategies/Strategy.php
```diff
@@ -24,6 +24,16 @@ public function __construct(DocumentationConfig $config)
         $this->config = $config;
     }
 
+    /**
+     * Returns a copy of the documentation config
+     *
+     * @return DocumentationConfig
+     */
+    public function getConfig()
```
shalvah, line 32: Curious, why did you add this method?

Also, it says it returns a copy of the config, but that isn't correct. It returns the same object.

src/Extracting/DatabaseTransactionHelpers.php
```diff
@@ -30,9 +49,60 @@ private function endDbTransaction()
 
         foreach ($connections as $conn) {
             try {
-                app('db')->connection($conn)->rollBack();
+                tap(
```
shalvah, line 52: Can we not use `tap()`? I understand its purpose, but I dislike it because it reads less clearly to me.

src/Extracting/DatabaseTransactionHelpers.php
```diff
@@ -30,9 +49,60 @@ private function endDbTransaction()
...
+                    app('db')->connection($conn),
+                    function ($driver) use ($conn) {
+                        if (self::driverSupportsTransactions($driver)) {
+                            return $driver->rollBack();
+                        }
+
+                        c::warn("Database driver [{$conn}] does not support transactions!");
```
shalvah, line 59: Terminology. We need to be consistent. Is it the connection or the driver that does not support transactions?🙂

config/scribe.php
Smudge3806: `use_without_transactions` sounds to me like we are introducing a secondary flow where no transactions will be used period.. perhaps I am nitpicking?! I agree that `allow_no_transactions` is a little confusing but 💯  that `use_without_transactions` is any less. I'll canvas for some opinions today with my team.

In terms of the nesting, this was actually where my thoughts where going at 3 AM this morning. In lue of a better name I 'll work this with the current name.

src/Extracting/Strategies/Strategy.php
Smudge3806: Because `DatabaseTransactionsHelper` is a trait that could be applied anywhere, and not exclusively to a child class of the abstract `Strategy`, the trait can't guarantee there is access to a instance property `$config`. Personally I think this is potentially something the language could add, however the correct approach at the trait level, would be to add the requirement for a method that would return an instance of the Config. Such method does not currently exist, so hence the new method.

Whilst I could have used in in-built `config` method from Laravel/Lumen, I felt you might prefer it if I stick to instances of the `DocumentationConfig`.

src/Extracting/DatabaseTransactionHelpers.php
Smudge3806: Yeah, `tap()` and `with()` appear to be relatively marmite in the reception. Consider it swapped out.

src/Extracting/DatabaseTransactionHelpers.php
Smudge3806: Hmmm... Honestly, It could be both, from the little reading around the subject I have done, some connections/databases don't support transactions and, library dependant, some drivers don't support transactions. Seen as an in-process transaction support **could** be developed for those drivers where the connections don't support transactions, I felt it prudent to lay the "blame" at the driver level rather than at the connection.

In the case of `Jenssegers`. MongoDB supports transactions, however Jenssegers has yet to implement this feature. It may also be the case that a user is using an older version of a library when the latest version does indeed support transactions. If that where the case, but the database itself was out-of-date and didn't support transactions, then surely that would be for the driver to handle?

Thoughts?

Smudge3806: Re-reading this, I think my logic is a bit squirrelly. Basically, I think there are a few things we need to consider:

1. Database supports transactions & driver does not (Most likely)
1. Database and driver do not support transactions (Not likely but possible)
1. Database does not support transactions but driver does (doesn't really affect us)

In terms of our messaging and terminology we have to decide where we lay the fault. As we connect to the database using a driver/adaptor/connector, I think it is prudent to place the fault at that level. 

Relating this to `Illuminate\Database`, the Transactions methods are added to the `Illuminate\Database\Connection` class via a Concern/trait. So the correct terminology in this case would be "Connection" however, if I know nothing about the inner workings of Laravel, would I consider "Connection" as the actual connection to the database server or the software responsible for initiating said connection?

In terms of design patterns, `Illuminate\Database` utilizes the Adaptor pattern for its various "Connection" classes. This would probably be the most correct terminology however, the subclasses of `Illuminate\Database\Connection` aren't stored or referred to as "Adaptors" so would this add confusion? Would the synonym "Driver" be more apt?

src/Extracting/Strategies/Strategy.php
Smudge3806: Fixed the description. Hopefully it's clearer.

config/scribe.php
Smudge3806: On review and consultation, we like `run_without_database_transactions` when in the array form.

src/Extracting/DatabaseTransactionHelpers.php
Smudge3806: I have updated the messaging to..
> Database driver for the connection [mongodb] does not support transactions!

This also maps to the message in the `DbTransactionSupportException` with.. 
> Database driver [Jenssegers\Mongodb\Connection] for connection [mongodb] does not support transactions. Changes to your database will be persistent. To allow this, add "mongodb" to the "run_without_database_transactions" config.

Thoughts?

shalvah: Yeah, that's good.

shalvah:
Fine with this, so I'm merging. A few minor nitpicks, but I'll add those in myself. Thanks!

src/Extracting/DatabaseTransactionHelpers.php
shalvah: Actually, I'm going to lay the blame on the driver. The connection is an arbitrary name the user picks.

shalvah:
Finally went with `continue_without_database_transactions`as the config name.😅 #57

PR #79, opened 2020-08-26
Feature/add context for route being processed

georgimorozov:
I think this would be a useful feature for those who need to contextually know what controller + action + middlware belongs to a route when documentation is being auto-generated using request classes. For instance, I dynamically generate bodyParams in request classes based on what middleware the route uses.

### DIFF
--- a/src/Extracting/Generator.php
+++ b/src/Extracting/Generator.php
@@ -32,7 +32,7 @@ class Generator
     }
 
     /**
-     * @return Route|null
+     * External interface that allows users to know what route is currently being processed
      */
     public static function getRouteBeingProcessed(): ?Route
     {
--- a/src/Extracting/Generator.php
+++ b/src/Extracting/Generator.php
@@ -20,12 +20,23 @@ class Generator
      */
     private $config;
 
-    public static $routeBeingProcessed = null;
+    /**
+     * @var Route|null
+     */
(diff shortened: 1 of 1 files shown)


shalvah:
Can you explain better, possibly with a code example? I don't really understand the use case.

georgimorozov:
Sorry I can't get into too much detail about it, but lets say you need access to the controller, action, and middleware associated with the route when calling bodyParameters() in a request class. If you `dd(route());` inside that method while generating docs, you will notice that you do not have the context of the correct action (and thus controller + middleware) which this request class is being called for. Things like Route::getCurrentRoute() and Route::currentRouteAction() dont work. I utilize the route-specific middleware and route name to populate some values in bodyParameters(), which is possible with this PR

src/Extracting/Generator.php
```diff
@@ -20,12 +20,25 @@ class Generator
...
+    private static $routeBeingProcessed = null;
+
     public function __construct(DocumentationConfig $config = null)
     {
         // If no config is injected, pull from global
         $this->config = $config ?: new DocumentationConfig(config('scribe'));
     }
 
+    /**
+     * @return Route|null
```
shalvah, line 35:
```suggestion
     * External interface that allows users to know what route is currently being processed
```

PR #84, opened 2020-09-01
Add ability to replace random token string in generated code examples

vr00:
This is about turning `"Authorization: Bearer ba5cdZPkVhaD4feg3v686E1"` into `"Authorization: Bearer {ACCESS_TOKEN}"`  in code examples, simply by setting the config option `auth.display_value` to `"{ACCESS_TOKEN}"`.  It would also replace random token value in code examples for other `auth.in` option values, not only `bearer`.

I also wrote short explanation in the docs, if you think something is missing, let me know. I'm open for suggestions.

P.S. I found no tests for `auth.display_value` config option, so I have no idea what and how should I test for. I would appreciate some help in that area.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -69,6 +69,12 @@ return [
         'use_value' => env('SCRIBE_AUTH_KEY'),
 
         /*
+         * The value of the parameter to be displayed by Scribe in documentation.
+         * If this value is null, Scribe will use a random value.
+         */
+        'display_value' => null,
+
+        /*
          * Any extra authentication-related info for your users. For instance, you can describe how to find or generate their auth credentials.
          * Markdown and HTML are supported.
          */
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -69,10 +69,10 @@ return [
         'use_value' => env('SCRIBE_AUTH_KEY'),
 
         /*
-         * The value of the parameter to be displayed by Scribe in documentation.
+         * Placeholder your users will see for the auth parameter in the example requests.
          * If this value is null, Scribe will use a random value.
          */
-        'display_value' => null,
+        'placeholder' => '{YOUR_AUTH_KEY}',
 
         /*
          * Any extra authentication-related info for your users. For instance, you can describe how to find or generate their auth credentials.
--- a/docs/config.md
+++ b/docs/config.md
@@ -102,6 +102,8 @@ Here are the available settings:
 
 - `use_value`: The value of the parameter to be used by Scribe to authenticate response calls. This will **not** be included in the generated documentation. If this value is null, Scribe will use a random value.
 
+- `display_value`: The value of the parameter that will be included in the generated documentation. If this value is null, Scribe will use a random value.
+
 - `extra_info`: Any extra authentication-related info for your users. For instance, you can describe how to find or generate their auth credentials. Markdown and HTML are supported. This will be included in the `Authentication` section.
 
 ### `strategies`
--- a/docs/config.md
+++ b/docs/config.md
@@ -102,7 +102,7 @@ Here are the available settings:
 
 - `use_value`: The value of the parameter to be used by Scribe to authenticate response calls. This will **not** be included in the generated documentation. If this value is null, Scribe will use a random value.
 
-- `display_value`: The value of the parameter that will be included in the generated documentation. If this value is null, Scribe will use a random value.
+- `placeholder`: Placeholder your users will see for the auth parameter in the example requests. If this is empty, Scribe will generate a realistic-looking auth token instead. Defaults to: "{YOUR_AUTH_KEY}".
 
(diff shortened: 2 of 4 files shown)


config/scribe.php
```diff
@@ -68,6 +68,12 @@
          */
         'use_value' => env('SCRIBE_AUTH_KEY'),
 
+        /*
+         * The value of the parameter to be displayed by Scribe in documentation.
+         * If this value is null, Scribe will use a random value.
+         */
+        'display_value' => null,
```
shalvah, line 75: Let's just make this default to something, so it's more obvious what it is.

Also, now that I think of it, `placeholder` would probably be more appropriate.

```suggestion
        'placeholder' => '{YOUR_AUTH_KEY}',
```

config/scribe.php
```diff
@@ -68,6 +68,12 @@
          */
         'use_value' => env('SCRIBE_AUTH_KEY'),
 
+        /*
+         * The value of the parameter to be displayed by Scribe in documentation.
```
shalvah, line 72:
```suggestion
         * Placeholder your users will see for the auth parameter in the example requests.
```

docs/config.md
```diff
@@ -102,6 +102,8 @@ Here are the available settings:
 
 - `use_value`: The value of the parameter to be used by Scribe to authenticate response calls. This will **not** be included in the generated documentation. If this value is null, Scribe will use a random value.
 
+- `display_value`: The value of the parameter that will be included in the generated documentation. If this value is null, Scribe will use a random value.
```
shalvah, line 105:
```suggestion
- `placeholder`: Placeholder your users will see for the auth parameter in the example requests. If this is empty, Scribe will generate a realistic-looking auth token instead. Defaults to: "{YOUR_AUTH_KEY}".
```

(I'm switching "if this is null" to "if this is empty" in the documentation generally, because it's both more technically correct and less of an implementation detail.)

docs/documenting/documenting-api-information.md
```diff
@@ -19,6 +19,7 @@ Here's how you'd configure auth with a query parameter named `apiKey`:
         'in' => 'query',
         'name' => 'apiKey',
         'use_value' => env('SCRIBE_API_KEY'),
+        'display_value' => 'YOUR-API-KEY',
```
shalvah, line 22:
```suggestion
        'placeholder' => '{YOUR_AUTH_KEY}',
```

docs/documenting/documenting-api-information.md
```diff
@@ -34,6 +35,7 @@ Here's an example with a bearer token (also applies to basic auth, if you change
         'in' => 'bearer',
         'name' => 'hahaha', // <--- This value is ignored for bearer and basic auth
         'use_value' => env('SCRIBE_AUTH_KEY'),
+        'display_value' => '{ACCESS-TOKEN}',
```
shalvah, line 38:
```suggestion
        'placeholder' => '{YOUR_AUTH_KEY}',
```

docs/documenting/documenting-api-information.md
```diff
@@ -47,13 +49,15 @@ And here's an example with a custom header:
         'in' => 'header',
         'name' => 'Api-Key', // <--- The name of the header
         'use_value' => env('SCRIBE_AUTH_KEY'),
+        'display_value' => 'YOUR-API-KEY',
```
shalvah, line 52:
```suggestion
        'placeholder' => '{YOUR_AUTH_KEY}',
```

src/Extracting/Generator.php
```diff
@@ -316,13 +316,14 @@ public function addAuthField(array $parsedRoute): array
         }
         $token = $faker->shuffle('abcdefghkvaZVDPE1864563');
         $valueToUse = $this->config->get('auth.use_value');
+        $valueToDisplay = $this->config->get('auth.display_value');
```
shalvah, line 319:
```suggestion
        $valueToDisplay = $this->config->get('auth.placeholder');
```

docs/documenting/documenting-api-information.md
```diff
@@ -47,13 +49,15 @@ And here's an example with a custom header:
...
         'extra_info' => 'You can retrieve your token by visiting your dashboard and clicking <b>Generate API token</b>.',
     ],
 ```
 
 You can set whatever you want as the `extra_info`. A good idea would be to tell your users where to get their auth key. 
 
 The `use_value` field is only used by Scribe for response calls. It won't be included in the generated output or examples.
+The `display_value` is opposite of `use_value`, it will be included in generated output in examples, but it will **not** be used for response calls.
```
shalvah, line 60: The `placeholder` is opposite of `use_value`, it will be used only as a placeholder in the generated example requests.

shalvah (changes requested):
Nice work. A few changes, but I'll implement those myself.

Don't worry about tests. You could write a small integration test (`GenerateDocumentationTest.php`) that verifies that the placeholder is indeed present in the output, but that's not a high-priority IMO. If you like, you can do it as a challenge and send in another PR for that,

PR #86, opened 2020-09-01
Generating form data with more complex arrays and objects

mnagy112:
Hi,

form data processing added in https://github.com/knuckleswtf/scribe/pull/13 does not count with more complex objects like this:

image

and results with error on creating documentation:

```
Facade\Ignition\Exceptions\ViewException 

  Array to string conversion (View: /Users/mnagy/AMCEF/Git/mapujme-new/backend/resources/views/vendor/scribe/partials/example-requests/bash.blade.php)

  at resources/views/vendor/scribe/partials/example-requests/bash.blade.php:12
     8| @endif
     9| @if(count($route['fileParameters']))
    10| @foreach($route['cleanBodyParameters'] as $parameter => $value)
    11| @foreach( \Knuckles\Scribe\Tools\WritingUtils::getParameterNamesAndValuesForFormData($parameter,$value) as $key => $actualValue)
  > 12|     -F "{!! "$key=".$actualValue !!}" \
    13| @endforeach
    14| @endforeach
    15| @foreach($route['fileParameters'] as $parameter => $file)
    16|     -F "{!! "$parameter=@".$file->path() !!}" @if(! ($loop->last))\@endif

      +2 vendor frames 
  3   [internal]:0
      Knuckles\Scribe\Writing\Writer::Knuckles\Scribe\Writing\{closure}()

  4   [internal]:0
      Knuckles\Scribe\Writing\Writer::Knuckles\Scribe\Writing\{closure}(Object(Illuminate\Support\Collection), "Maps")
```

WritingUtils and PostmanWriter were changed to remove error and create correct postman collection:

image

If there are any problems with PR, just comment out, I will try to fix them, contributing first time :)

### DIFF
--- a/src/Tools/WritingUtils.php
+++ b/src/Tools/WritingUtils.php
@@ -111,23 +111,35 @@ class WritingUtils
      *
      * @return array
      */
-    public static function getParameterNamesAndValuesForFormData(string $parameter, $value): array
+    public static function getParameterNamesAndValuesForFormData($parameter, $value)
     {
-        if (!is_array($value)) {
-            return [$parameter => $value];
+        $results = [];
+
+        if (is_array($value) && ! empty($value)) {
+            foreach ($value as $key => $v) {
+                if (is_numeric($key)) {
+                    $key = '';
+                }
+
+                if (is_array($v) && ! empty($v)) {
+                    $results = array_merge($results, static::getParameterNamesAndValuesForFormData($parameter.'['.$key.'][', $v));
+                } else {
+                    if ($v != null) {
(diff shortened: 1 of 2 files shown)


mnagy112:
There are more issues with files in form data request and these more complex request bodies. Changes are ready but need to be tested properly and cleaned up. New PR will be opened then.

shalvah:
Cool. You may want to hold on a bit, though. I made some improvements to the Collection generation, and I'm introducing a clearer syntax for arrays and objects that eliminates most of the confusion. Should be up in a few days. I'll tag you when it is.

PR #89, opened 2020-09-03
Properly iterate over all database transactions

wrurik:
I ran into problems where generating documentation would result in data being added to my database.
After investigating I found that a DatabaseTransaction is only started for the first connection in the config.

[This comment](https://github.com/knuckleswtf/scribe/pull/5#pullrequestreview-407382542) on a previous PR makes me believe the intention is to start a transaction for all available connections. 
This PR should fix that


shalvah (commented):
Ouch, that's a nasty one. Nice catch

shalvah:
Will tag a new release later today

PR #93, opened 2020-09-08
Added Support for Laravel 8.x

Jurager:
Edited composer.json and composer.dingo.json. Tests in local environment passed. Went through the Laravel 8.x migration documentation, it looks like no more changes are required


shalvah:
Did you test on an 8.x application? I believe some backwards-incompatible changes were made to model favorites?

shalvah:
Also noticed that the CI job installs Laravel 7 on the "highest" test, so it really hasn't been tested on 8. https://travis-ci.com/github/knuckleswtf/scribe/jobs/382168472

Needs to be a way to force that.

Jurager:
> Did you test on an 8.x application? I believe some backwards-incompatible changes were made to model favorites?

Yep, i've tested it on local installation of 8.x-dev, all seems to be fine with models, cause i can generate docs with sample responses from database

> Also noticed that the CI job installs Laravel 7 on the "highest" test, so it really hasn't been tested on 8. https://travis-ci.com/github/knuckleswtf/scribe/jobs/382168472

I didn't see "highest" test at all, only stable, but i can suggest that, it didn't tested, cause it uses composer --prefer-stable, when actually test was processed there is no stable version of 8.x, and it used last stable 7.x

For now, i see that stable 8.0.0 was released, maybe need to re-run the test?

shalvah:
Will merge this soon. Just need to make sure the Travis builds actually run on an 8.x instance. The highest it gets is 7.28.

composer.json
```diff
@@ -36,7 +36,7 @@
   "require-dev": {
     "brianium/paratest": "^4.0",
     "dms/phpunit-arraysubset-asserts": "^0.1.0",
-    "laravel/lumen-framework": "^5.8|^6.0|^7.0",
+    "laravel/lumen-framework": "^5.7|^6.0|^7.0|^8.0",
```
shalvah, line 39: I'm guessing this was a merge error or something.

``suggestion
    "laravel/lumen-framework": "^5.8|^6.0|^7.0|^8.0",
```

shalvah:
Ah, seen the problem. Lumen 8 isn't released yet (few days' time), and it's part of our dev dependencies. I'll update it when it's released.

fcolecumberri:
> Ah, seen the problem. Lumen 8 isn't released yet (few days' time), and it's part of our dev dependencies. I'll update it when it's released.

do you know when Lumen 8 will be  released?

shalvah:
> Illuminate\Contracts\Container\BindingResolutionException: Target class [Illuminate\Database\Eloquent\Factory] does not exist.

Yup, I was right. It fails on Laravel 8.

shalvah:
I'll fix it and tag the L8-compatible release soon. Won't work on Lumen though. Lumen 8 not out yet.

shalvah:
Tagged 1.7.0, which should support installation with L8.

PR #94, opened 2020-09-10
More flexible responses file location

sterxa:
So, because we are using Envoyer and it skips the whole storage folder on deployment we had to slightly modify this great package. Now you can set the base response files dir in the configuration.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -118,6 +118,8 @@ INTRO
 
     'description' => '',
 
+    'responses_base_dir' => storage_path('responses/'),
+
     /*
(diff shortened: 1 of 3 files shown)


shalvah:
Hmmm I hate adding new config keys. It leads to an overly bloated file. I'm thinking we just make the response file path a path relative to your project root instead.

shalvah:
Thanks for the contribution. Sorry, I have to turn it down though. We'll switch to storage-independent paths (absolute or relative to your project directory) in v2, and it's easier for me to just implement that now.

PR #96, opened 2020-09-14
Fix openapi route

### LINKED ISSUE #95: OpenAPI link within docs shows empty object
Hey @shalvah , sorry for not following the template. I know exactly what the issue is here.

In the generated docs, we get a link: ``View OpenAPI (Swagger) Spec``.

If we click that, we get taken to ``\Knuckles\Scribe\Http\Controller::openapi``

Then there is an attempt to respond in the same way as the postman json collection, by json decoding a yaml file, and then re-encoding it as a json response.

This leads to an empty json object as a response, when in fact the underlying ``openapi.yaml`` file does exist and is populated properly.

A couple of solutions here:

1) We use Symfony's Yaml component to first convert the yaml to php array, then give the response as json, like so:

```
use Symfony\Component\Yaml\Yaml;

class Controller
{
...
public function openapi()
    {
        return response()->json(
            Yaml::parse(
                Storage::disk('local')
                       ->get('scribe/openapi.yaml')
            )
        );
    }
}
```

2) We convert this endpoint into a file download endpoint instead of a json endpoint, like so:

```
    public function openapi()
    {
        return response()->download(storage_path('app/scribe/openapi.yaml'), 'openapi.yaml', ['Content-Type' => 'application/x-yaml']);
    }
```

3) For consistency we convert both postman and openapi endpoints to file download types.

I am happy to submit a PR if you have a preferred approach.

Kind regards,

Rob

robtesch:
Fixes #95

### DIFF
--- a/src/Http/Controller.php
+++ b/src/Http/Controller.php
@@ -23,10 +23,13 @@ class Controller
         );
     }
 
+    /**
+     * @throws \Illuminate\Contracts\Filesystem\FileNotFoundException
+     *
+     * @return \Symfony\Component\HttpFoundation\BinaryFileResponse
+     */
     public function openapi()
     {
-        return response()->json(
-            json_decode(Storage::disk('local')->get('scribe/openapi.yaml'))
-        );
+        return response()->file(Storage::path('scribe/openapi.yaml'));
     }
 }


shalvah:
Thanks.

shalvah:
This will output the file content, right?

robtesch:
> This will output the file content, right?

That's correct yes, it will just output the file content in the browser as requested.

Issue #56, opened 2020-07-09
Models created by factory are deleted before calling the resource


PollieDev:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.readthedocs.io/en/latest/troubleshooting.html).**

**What happened?**
*This is is what I assume is the problem, I could be wrong. Apologies if that's the case.*

Some endpoints of ours have a model which creates multiple relations with the use of factories.
In our case we have an application which belongs to a stage. Which then again belong to a job.
So in our application factory we create a stage if one is not defined.
And in the stage factory we create a job if one is not defined.

The creation of the items are working perfectly but once the model (the application) hits our resource the job doesn't exist anymore. The stage does still exist and works perfectly. 

Example code:

Application Factory
```php
$factory->define(Application::class, function (Faker $faker, $params) {
    if (isset($params['stage_id'])) {
        $stage = Stage::findOrFail($params['stage_id']);
    } else {
        $stage = factory(Stage::class)->create();
    }

    $stage->job // exist just as needed!

    return [
        // Attributes
    ];
});
```

Stage Factory
```php
$factory->define(Stage::class, function (Faker $faker, $params) {
    if (isset($params['job_id'])) {
        $job = Job::with('company.jobsite')->findOrFail($params['job_id']);
    } else {
        $job = factory(Job::class)->create();
    }

    $job // Job exist perfectly fine

    return [
        // Attributes
    ];
});
```

Resource
```php
class Application extends JsonResource
{
    /**
     * Transform the resource into an array.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return array
     */
    public function toArray($request): array
    {
         $this->stage // exist just as expected.
         $this->stage->job_id // also exist and has the ID that was provided in the factory
         $this->stage->job // does not exist. We use this relation on many places so we are pretty sure it's not the relation
         Job::where('id', $this->stage->job_id)->first() // + this also returns null
    }
}
```

This is the docblock on the controller (if you'd need that information):
```php
    /**
     * Retrieve an application
     *
     * @header X-Application-Access-Token {access_token}
     * @apiResource \App\Http\Resources\Content\Application
     * @apiResourceModel \App\Models\Application
     *
     * @param integer $id
     * @return \App\Http\Resources\Content\Application
     */
```

So it looks like the job has already been removed before I could make use of it.

**My environment:**
 - PHP version: 7.4.7
[...]

Issue #46, opened 2020-06-19
Load relations not working


ilgala:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.readthedocs.io/en/latest/troubleshooting.html).**

**What happened?**
I'm trying to create the documentation for a listing endpoint where the returned collection models should also load some relations, here the doc block:

```php
/**
 * My model - index
 * bla bla, bla bla bla bla, bla.
 *
 * @group My Group
 * @queryParam XXX required description. Example: "XXX"
 * @queryParam YYY description. Example: "YYY"
 * @apiResourceCollection \My\Resource\Collection
 * @apiResourceModel \My\Resource\Model paginate=10 states=docs with=rel1,rel2,rel3
 *
 * @param \Illuminate\Http\Request $request
 *
 * @return \My\Resources\Collection
 */
public function index(Request $request) {
  // MY CODE
}
```

I expect something like:

```JSON
{
  "data": [
    [{
      "param": "xxx", 
      "param": "yyy", 
      "rel1": {
        "rel_param": "zzz"
      },
      "rel2": {
        "rel_param": "zzz"
      },
    }],
    // THE OTHER RESULTS
  ]
}
```
but in the generated documentation I only see:
```JSON
{
  "data": [
    [{
      "param": "xxx", 
      "param": "yyy"
    }],
    // THE OTHER RESULTS
  ]
}
```

**My environment:**
 - PHP version: 7.4.7
 - Laravel version: 6.18.20
 - Scribe Version: 1.1.0

**My Scribe config:**
```php
return [
    'type' => 'laravel',
    'static' => [
        'output_path' => 'public/docs',
    ],
    'laravel' => [
        'add_routes' => false,
        'docs_url' => '/docs',
        'middleware' => [],
    ],
    'auth' => [
        'enabled' => true,
        'in' => 'bearer',
        'name' => 'Authorization',
        'use_value' => env('SCRIBE_AUTH_KEY'),
        'extra_info' => 'You can retrieve your token by visiting your dashboard and clicking <b>Generate API token</b>.',
    ],
    'intro_text' => <<<'INTRO'
Welcome to DAPL API documentation!

<aside>As you scroll, you'll see code examples for working with the API in different programming languages in the dark area to the right (or as part of the content on mobile), and you can switch the programming language of the examples with the tabs in the top right (or from the nav menu at the top left on mobile).</aside>
INTRO
    ,
    'example_languages' => [
[...]

PR #104, opened 2020-09-24
Issues 56 and 46 fix proposal

ilgala:
✅ This is a fix proposal for the bugs related to the issues:

* [Models created by factory are deleted before calling the resource](https://github.com/knuckleswtf/scribe/issues/56)
* [Load relations not working](https://github.com/knuckleswtf/scribe/issues/46)

🛠️ Major changes
* Database transactions moved into `getApiResourceResponse` to avoid relation deletion before a lazy load.
* Added the route resolver

### DIFF
--- a/tests/Strategies/Responses/UseApiResourceTagsTest.php
+++ b/tests/Strategies/Responses/UseApiResourceTagsTest.php
@@ -2,11 +2,13 @@
 
 namespace Knuckles\Scribe\Tests\Strategies\Responses;
 
+use Illuminate\Routing\Route;
 use Knuckles\Scribe\Extracting\Strategies\Responses\UseApiResourceTags;
 use Knuckles\Scribe\ScribeServiceProvider;
 use Knuckles\Scribe\Tests\Fixtures\TestUser;
 use Knuckles\Scribe\Tools\DocumentationConfig;
 use Knuckles\Scribe\Tools\Utils;
+use Mockery;
 use Mpociot\Reflection\DocBlock\Tag;
 use DMS\PHPUnitExtensions\ArraySubset\ArraySubsetAsserts;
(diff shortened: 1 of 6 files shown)


tests/Strategies/Responses/UseApiResourceTagsTest.php
```diff
@@ -77,12 +86,18 @@ public function can_parse_apiresource_tags_with_model_factory_states()
     {
         $config = new DocumentationConfig([]);
 
+        $route = Mockery::mock(Route::class);
```
shalvah, line 89: Can you add a comment explaining why we have this?

shalvah (changes requested):
Can you implement the same for the Transformer strategy too?

tests/Strategies/Responses/UseApiResourceTagsTest.php
ilgala: Because if you look at these two Api Resources Knuckles\Scribe\Tests\Fixtures\TestUserApiResource (lines 27-29) and Knuckles\Scribe\Tests\Fixtures\TestUserApiResourceCollection (lines 25-27). They now call the $request->route() method:

if($request->route()->named('test')) {
    $data['test'] = true;
}
I thought that mocking the request was a good way to test that if your logic includes calling the $request object inside an APIResource, the package instantiate the resource with all the necessary data.

ilgala:
> Can you implement the same for the Transformer strategy too?

Added on last commit

tests/Strategies/Responses/UseApiResourceTagsTest.php
shalvah: I meant adding a comment in the code, so the next person who sees it can know, but I'll handle that.

PR #114, opened 2020-10-14
Set correct base URL protocol for Postman collection

wjonkerhulst:
The protocol cloud not be correctly resolved and would always default to http. This because $baseUrl would always contain the url witouth http or htpps prefix. Because of this the function `Str::startsWith($baseUrl, 'https') ? 'https' : 'http'` will always return http.

```php
$baseUrl = $this->getBaseUrl($this->config->get('postman.base_url', $this->config->get('base_url')));
$base = [
          'protocol' => Str::startsWith($baseUrl, 'https') ? 'https' : 'http',
];
```

Change the code as bellow. This will resolve the correct prefix for the url (eithet http or https)
```php
$protocol = Str::startsWith($this->config->get('postman.base_url', $this->config->get('base_url')), 'https') ? 'https' : 'http';
$baseUrl = $this->getBaseUrl($this->config->get('postman.base_url', $this->config->get('base_url')));
$base = [
            'protocol' => $protocol,
];
```

### DIFF
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -200,9 +200,10 @@ class PostmanCollectionWriter
             return Str::contains($route['uri'], '{' . $key . '}');
         });
 
+        $protocol = Str::startsWith($this->config->get('postman.base_url', $this->config->get('base_url')), 'https') ? 'https' : 'http';
         $baseUrl = $this->getBaseUrl($this->config->get('postman.base_url', $this->config->get('base_url')));
         $base = [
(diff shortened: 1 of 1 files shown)


shalvah:
Nice catch. Why is the build failing though? It's complaining about a syntax error on PHP 7.3. https://travis-ci.com/github/knuckleswtf/scribe/jobs/399681690

shalvah:
(I suspect it may be a bug in an external dependency. Rerunning the build - https://travis-ci.com/github/knuckleswtf/scribe/builds/190064973)

wjonkerhulst:
Thanks!

shalvah:
Released in 1.9.1

PR #125, opened 2020-10-27
Allow queryParam array paramter for tryout

styler2go:
Previously there was a bug where executing an "tryitout"-request with an queryParameter like fields[users]=name was not working. It executed it as `fields=[object Object]`. To fix this, i created another Loop which resolves these objects and adds them to the searchParams.

### DIFF
--- a/resources/js/tryitout.js
+++ b/resources/js/tryitout.js
@@ -51,6 +51,17 @@ function cancelTryOut(endpointId) {
     }
 }
 
+function recursiveAddSearchParams(url, previousAppend, previousKey, obj) {
+    if (typeof obj === 'object') {
+        Object.keys(obj).forEach(objKey => {
+            const nextParam = !previousKey ? previousAppend : `${previousAppend}[${previousKey}]`;
+            recursiveAddSearchParams(url, nextParam, objKey, obj[objKey]);
+        });
+    } else {
+        url.searchParams.append(`${previousAppend}[${previousKey}]`, obj);
+    }
+}
+
 function makeAPICall(method, path, body, query, headers) {
     console.log({path, body, query, headers});
 
@@ -61,9 +72,7 @@ function makeAPICall(method, path, body, query, headers) {
     const url = new URL(window.baseUrl + '/' + path.replace(/^\//, ''));
     Object.keys(query).forEach(key => {
         if (typeof query[key] === 'object') {
-            Object.keys(query[key]).forEach(objKey => {
-                url.searchParams.append(`${key}[${objKey}]`, query[key][objKey]);
-            });
+            recursiveAddSearchParams(url, key, null, query[key]);
         } else {
             url.searchParams.append(key, query[key]);
         }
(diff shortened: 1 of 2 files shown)


shalvah:
Haha, I fell for the same mistake I described in my article: https://dev.to/shalvah/fun-stuff-representing-arrays-and-objects-in-query-strings-36bn

I think the PR needs some more work. As it is, it's only going to handle one level of nesting. How about extracting that into a function so we can just use recursion?

styler2go:
That sounds like a good extension to the existing code. I will look into it and update the PR

styler2go:
@shalvah What do you think about this code? Is it clean enough for you for the tryitout.js?

https://jsfiddle.net/brzg05nx/1/

resources/js/tryitout.js
```diff
@@ -51,6 +51,17 @@ function cancelTryOut(endpointId) {
     }
 }
 
+function recursiveAddSearchParams(url, previousAppend, previousKey, obj) {
```
shalvah, line 54: Let's make this sound more natural. `addQueryParamsRecursively` is fine.

shalvah, line 54: Also, `previousAppend, previousKey, obj` -> `parent, key, value`. More relevant names. Also, recursion typically does not care about "previous" values. What it needs is the data that applies to the "current" run.

resources/js/tryitout.js
```diff
@@ -61,9 +72,7 @@ function makeAPICall(method, path, body, query, headers) {
     const url = new URL(window.baseUrl + '/' + path.replace(/^\//, ''));
     Object.keys(query).forEach(key => {
         if (typeof query[key] === 'object') {
-            Object.keys(query[key]).forEach(objKey => {
-                url.searchParams.append(`${key}[${objKey}]`, query[key][objKey]);
-            });
+            recursiveAddSearchParams(url, key, null, query[key]);
```
shalvah, line 75: We don't need the outer `Object.keys` loop anymore. We can simply call the recursive function right after the `const url` line. It should work, since the recursive function also handles non-object values.

resources/js/tryitout.js
shalvah, line 54: Also no need to pass in the full `url`. `url.searchParams` is what we need, so you can change the parameter to `searchParams`.

resources/js/tryitout.js
```diff
@@ -51,6 +51,17 @@ function cancelTryOut(endpointId) {
     }
 }
 
+function recursiveAddSearchParams(url, previousAppend, previousKey, obj) {
+    if (typeof obj === 'object') {
+        Object.keys(obj).forEach(objKey => {
+            const nextParam = !previousKey ? previousAppend : `${previousAppend}[${previousKey}]`;
```
shalvah, line 57: This is pretty confusing. In what scenario would there not be a key (aka `previousKey`)?

shalvah (changes requested):
Nice work, but need some changes.

shalvah:
Hey @styler2go any progress on this?

shalvah:
Hey @styler2go I went ahead to implement this. Thanks for your hard work.

Issue #120, opened 2020-10-22
Slow to generate docs


rmundel:
The verbose options aren't give me any insights (everything runs without errors) on why my documentation is taking very long to generate.
I have only 3 api routes and is taking like 10 minutos or so.
Any tips on where to start debugging it?

shalvah:
What version are you on? Try v2 beta and see if there are any improvements.

I have speed improvements planned for v3, but that won't be until next year. In the meantime, you can also try disabling strategies you don't need by commenting them out in the config file. For instance, if you don't use Fractal Transformers, you can comment out the transformer tags strategy.

rmundel:
I'm on 1.9.1. Just tried disabling every strategy there is, and the time was the same. Gonna to try v2.

shalvah:
Also, are you limiting your config to your API routes only? Sometimes people forget.

tjep82:
Had the same issue, i commented the following in the config strategies:

```
'bodyParameters' => [
            //\Knuckles\Scribe\Extracting\Strategies\BodyParameters\GetFromFormRequest::class,
           \Knuckles\Scribe\Extracting\Strategies\BodyParameters\GetFromBodyParamTag::class,
        ],
        'responses' => [
            //\Knuckles\Scribe\Extracting\Strategies\Responses\UseTransformerTags::class,
            \Knuckles\Scribe\Extracting\Strategies\Responses\UseResponseTag::class,
            \Knuckles\Scribe\Extracting\Strategies\Responses\UseResponseFileTag::class,
           //\Knuckles\Scribe\Extracting\Strategies\Responses\UseApiResourceTags::class,
           // \Knuckles\Scribe\Extracting\Strategies\Responses\ResponseCalls::class,
```

These are the strategies that is causing the slow process try that.

R0ad-Runn3r:
Hey, I had the same issue but just inside my docker. Commenting these lines, @tjep82 suggested, "solved" it, but i don't think that this is the solution. I am actually not able to provide a solution, do you have any ideas?

aaronflorey:
I've had the same issue and found it's because of a change that was made recently, that would create a transaction for every single database in config/database.php 

We're using the default database.php config file which includes mssql, pgsql, mysql and sqlite - so this package is trying to connect to every database and start a transaction to that database.

Even worse is that i don't have drivers for mssql or pgsql, so it was causing a "pdo driver not found" error, and then the other errors were "can't connect to server"

[...]

PR #138, opened 2020-11-05
only loop over enabled database connections

aaronflorey:
We had an issue where we had multiple databases configured in config/database.php - Scribe was looping over each database creating a transaction for this which was causing slow downs and exceptions.

I opted to change continue_without_database_transactions into enable_transactions_for_database_connections where the user can list all database connections that require a transaction, defaulting to whatever database is default.

Our biggest issue was that we use MSSQL in production, but MySQL in testing. So scribe would throw exceptions because the MSSQL driver wasn't installed on our docker image, but this also caused each endpoint to take 10m to respond. Very strange.

Thanks

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -362,7 +362,11 @@ INTRO
     /**
      * [Advanced] If one of your app's database drivers does not support transactions,
      * docs generation (instantiating Eloquent models and making response calls) will likely fail.
-     * To avoid that, you can add the driver class name here. Be warned: that means all database changes will persist.
+     * We loop over all of your database connections and start a transaction if supported.
+     * Here you can list all of the connections you want us to start a transaction on.
+     * Typically you would list the name of every connection your database will connect to while running this command.
      */
-    'continue_without_database_transactions' => [],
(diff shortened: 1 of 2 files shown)


shalvah:
+1 on the attempt to limit it to "enabled" connections. However, there are a number of issues here:
- It's a breaking change, so can't come in in v2.
- It's a "destructive" approach.  `enable_transactions_for_database_connections` would merely fail that part of the docs with a warning if you forgot to set it. But `enable_transactions_for_database_connections` as you have it now, might end up writing data to someone's database if they forgot to set it. 
- Also, you can't really define "enabled" connections in Laravel. If it's in your `config/database.php`, it's enabled, and ou can switch to it at any time by specifying the driver.

I'm thinking we can just go with https://github.com/knuckleswtf/scribe/issues/120#issuecomment-721910667 instead.🤔

shalvah:
Hi @aaronflorey thanks for the effort, but I've gone ahead to fix the speed issue in a different way. Kept things as is and used amphp/parallel0functions to launch the transactions in parallel. Test it out and let me know how it works.

PR #149, opened 2020-11-30
Fixed throwing exception during fetching connection driver for DB transactions.

milosh012:

### DIFF
--- a/src/Extracting/DatabaseTransactionHelpers.php
+++ b/src/Extracting/DatabaseTransactionHelpers.php
@@ -14,30 +14,28 @@ trait DatabaseTransactionHelpers
 {
     private function startDbTransaction()
     {
-        $connections = array_keys(config('database.connections', []));
+        $connection = config('database.default', 'mysql');
 
-        wait(parallelMap($connections, function ($connection) {
-            try {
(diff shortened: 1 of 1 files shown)


shalvah:
Thanks for the effort. This really isn't the fix, though. It's a change of behaviour (something of a regression to older behaviour). For instance, the current version attempts to start a transaction for each database connection, while your proposed change only does so for the default.

But I'll merge it because it will save me a bit of work in implementing the fix.

milosh012:
@shalvah yeah, its not really a fix... but I dont get it why you need to loop over each connection? DB is  in most of the cases on one connection (default). I dont have so much exp with AMP, but the problem is there - it can not fetch DB application (maybe we can also get rid of AMP completely now because its not used anymore)

shalvah:
> DB is in most of the cases on one connection (default).

True, but not always. And for those folks who use more than one connection, not starting transactions in all connections would be surprising behaviour right now. At the very least, we'd need to change behaviour so Scribe screams the warning.

.>  (maybe we can also get rid of AMP completely now because its not used anymore)
Once I've decided on the fix.🙂

PR #195, opened 2021-04-07
If param name contains password set input type to password

rubenpoppe:
If the param name contains 'password' hide the input text by setting `type="password"` for `input`.

Before:
image
After:
image


shalvah:
Smart move!

PR #198, opened 2021-04-09
format form data params properly

aaronflorey:
This PR will format the form params properly in Postman.

Currently an object would be formatted eg. patient = [Object object] in Postman

This PR corrects that by formatting it as patient[key] = value instead


shalvah:
Good work.👍

PR #203, opened 2021-04-27
Add ability to override the default parameter type.

JoeBocock:
Tiny PR to add the option to override the default parameter type of `string`.

I couldn't see a way to do this out of the box and I looked over the documentation for it. Really sorry if I've just completely missed it 😅 

I'm aware that config clutter is a pain so please feel free to tear this apart. I'd be happy to go away and implement it a completely different way if needed!

I wasn't really sure what tests I could write for this but if you'd like me to add a specific one, let me know.

There's a tiny documentation change so as requested:

Before
Before

After
After

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -307,6 +307,17 @@ INTRO
     'faker_seed' => null,
 
     /**
+     * If Scribe cannot infer the type required for a parameter, it will default to string.
+     * Change this value if you'd like to adapt this behaviour.
+     *
+     * Available values:
+     *  - 'string'
+     *  - 'integer'
+     *  - 'number'
+     */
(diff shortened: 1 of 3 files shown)


shalvah:
Glad you recognize the problem with config clutter, but still...this is. Defaulting to `string` seems very sensible to me, not worthy of configurability. After all, it's pretty easy to add an `int` or `bool` to a description. If you really can't do that, you could always create your own custom strategy and use that instead.

JoeBocock:
That's a shame!

While I disagree, I understand. Thanks for taking a look.

shalvah:
All good.👍 If you want custom behaviour, you can subclass the strategy and replace the default one in your config.

PR #207, opened 2021-05-05
unescape unicode when generate postman collections for other locale

guitarbien:
In some locales like **zh-tw**, `json_encode()` will make the content to `\u....`
after_postman

I want to add the flag `JSON_UNESCAPE_UNICODE` could show the correct words, and will not affect other locales
after_postman

Thanks.


shalvah:
Good PR, thank you.

Issue #189, opened 2021-03-04
FormRequest Dependency Injection


eac0n:
According to the [Laravel docs](https://laravel.com/docs/8.x/validation#creating-form-requests) you can inject dependencies to the` rules` method of a `FormRequest`.
> You may type-hint any dependencies you require within the rules method's signature. They will automatically be resolved via the Laravel service container.

However, scribe seems to have trouble with this since the `GetFromFormRequest` strategy always calls the `rules` method without parameters. I managed to workaround the Problem by overwriting `GetFromFormReques::getRouteValidationRules` in a custom strategy like this:

```php
namespace App\Docs\Strategies\BodyParameters;


use Illuminate\Contracts\Validation\Factory as ValidationFactory;
use Illuminate\Support\Facades\App;

class GetFromFormRequest extends \Knuckles\Scribe\Extracting\Strategies\BodyParameters\GetFromFormRequest {

    protected function getRouteValidationRules($formRequest)
    {
        if (method_exists($formRequest, 'validator')) {
            $validationFactory = app(ValidationFactory::class);

            return call_user_func_array([$formRequest, 'validator'], [$validationFactory])
                ->getRules();
        } else {
            $params = [];
            $reflectionMethod = new \ReflectionMethod(get_class( $formRequest ), 'rules');
            $reflectionParameters = $reflectionMethod->getParameters();
            foreach($reflectionParameters as $reflectionParameter) {
                $params[] = App::make($reflectionParameter->getClass()->getName());
            }
            return call_user_func_array([$formRequest, 'rules'], $params);
        }
    }

}
```
I don't know if this will work for every dependency but in my case, a custom rule, it works fine. Also I don't know if other strategies may face the same problem.

Nevertheless, I think scripe should take dependency injection into account of its own accord.

shalvah:
Yes, the form request strategy always instantiates the form request directly, not through the service container.

The FormRequest behaviour is complicated. We could change it to use the service container, but that won't fix every thing. 🤔

tbordinat:
Hi!
Thank you for your awesome work on this project !
I'm confronted to something similar, do you know if there is a solution ? I looked for one but I can't find anything... :/ 

Here is my request.

```
class AnalyticsRequest extends FormRequest
{
    public function __construct(
        UserService $userService
    ) {
        $this->userService = $userService;
    }
    /** ... */
}
```

And I got this error when I try to `scribe:generate` on my project : 

```
[...]

PR #217, opened 2021-05-21
Skip FormRequest if instanciation isn't possible

tbordinat:
Hi @shalvah !

I don't know if this is acceptable but I found a solution to my problem and I think it can help some of us :)

In case FormRequest class needs dependency injection, instanciation throws an ArgumentCountError exception.
By catching it, we can simply skip this class and the developer can use `@bodyParam` annotation to add documentation on the params.

See https://github.com/knuckleswtf/scribe/issues/189

### DIFF
--- a/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
+++ b/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
@@ -2,6 +2,7 @@
 
 namespace Knuckles\Scribe\Extracting\Strategies\BodyParameters;
 
+use ArgumentCountError;
 use Dingo\Api\Http\FormRequest as DingoFormRequest;
 use Illuminate\Foundation\Http\FormRequest as LaravelFormRequest;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
```diff
@@ -66,8 +67,14 @@ public function getBodyParametersFromFormRequest(ReflectionFunctionAbstract $met
             if (
                 (class_exists(LaravelFormRequest::class) && $parameterClass->isSubclassOf(LaravelFormRequest::class))
                 || (class_exists(DingoFormRequest::class) && $parameterClass->isSubclassOf(DingoFormRequest::class))) {
-                /** @var LaravelFormRequest|DingoFormRequest $formRequest */
-                $formRequest = new $parameterClassName;
+                try {
+                    /** @var LaravelFormRequest|DingoFormRequest $formRequest */
+                    $formRequest = new $parameterClassName;
+                } catch (ArgumentCountError $e) {
+                    c::info('Skipping instanciation of ' . $parameterClassName . ' because of dependency injection. Use manual @bodyParam to describe this request.');
```
shalvah, line 74: Spelt with a _t_, not _c_.

```suggestion
                    c::info('Skipping instantiation of ' . $parameterClassName . ' because of dependency injection. Use manual @bodyParam to describe this request.');
```

shalvah (approved):
I'm not sure if this is the direction I'd prefer to go, but I guess it's better than crashing in the meantime.

Issue #212, opened 2021-05-20
Advanced Form Request rules crash the documentation 


andronikigr:
Hello! Thanks for the amazing material!

I want to ask if is there any way to overcome some obstacles:
I have written some advanced rules in an UpdateSomehtingRequest for nested objects.

I also have registered some custom validation rules.

One rule is something like this:

'data.*.results'            => ['required', 'array', ['results.requiredGroupIds', $this->thing]],

So in this rule, scribe crashes out. 
I want to ask if is there a way to get around that manually.
I tried to specify the bodyparameter for that rule but I get the same error:

🚫 error ErrorException in /var/www/html/vendor/knuckleswtf/scribe/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php at line 408: strpos() expects parameter 1 to be string, array given.

The thing is that we have many custom rules! 
So we will be more than thankful if is there a solution!

andronikigr:
@shalvah 
Thank you for your response!
This is something that we've used many times and it has worked fine!
It is an alternative way when you want to pass something that is not scalar variable.

But we have this exception while trying to generate documentation through scribe, 
so I ask  if there is any solution to avoid the problem in some way.

Thanks a lot in advance!

shalvah:
Obviously, this is a bug in Scribe that should be fixed. At worst, the extra rules should be ignored. Just need to be sure that that is valid Laravel syntax. I'm not aware of array rules. Any links?

shalvah:
I'll just add a defensive check that checks the type of the valuation rule first. I'm busy now, so you can send in a PR for that; it's easy.

Just add an if-condition to this foreach https://github.com/knuckleswtf/scribe/blob/cf9d9765ac4fd48e4bc02d18df22411ca2cee1de/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php#L197 that does a `continue` if the type of `$rule` is not string or `Rule` object.

andronikigr:
@shalvah 
Thanks again!

I've sent a PR , as you asked:
https://github.com/knuckleswtf/scribe/pull/219

WesleyGoncalves:
> Obviously, this is a bug in Scribe that should be fixed. At worst, the extra rules should be ignored. Just need to be sure that that is valid Laravel syntax. I'm not aware of array rules. Any links?

It can be found on the following section of the documentation [Validating Arrays](https://laravel.com/docs/8.x/validation#validating-arrays)

[...]

PR #219, opened 2021-05-24
Ignore rules other than `string` rules or `Rule` objects

andronikigr:
This Pull Request will resolve the issue here: https://github.com/knuckleswtf/scribe/issues/212

@shalvah 
I follow the suggestions you made in your answer to my question.

### DIFF
--- a/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
+++ b/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
@@ -196,6 +196,7 @@ class GetFromFormRequest extends Strategy
         $testData = [];
         foreach ($rules as $key => $ruleset) {
             if (!Str::contains($key, '.*')) continue;
+            if (!is_string($ruleset) && !($ruleset instanceof Rule)) continue;
 
             // All we need is for Laravel to see this key exists
             Arr::set($testData, str_replace('.*', '.0', $key), Str::random());


src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php
```diff
@@ -196,6 +196,7 @@ protected function normaliseRules(array $rules)
         $testData = [];
         foreach ($rules as $key => $ruleset) {
             if (!Str::contains($key, '.*')) continue;
+            if (!is_string($ruleset) && !($ruleset instanceof Rule)) continue;
```
shalvah, line 199: Did you test this? I think the `ruleset` here is actually the group of all rules, which should be an array. I believe this should go in `parseRule()`, the function that parses each rule individually. Even better, this method (where therre's already a check for `Rule` class):

https://github.com/knuckleswtf/scribe/blob/73351b170450d69e4039edd58034a883d4b0648e/src/Extracting/Strategies/BodyParameters/GetFromFormRequest.php#L399-L409

shalvah (changes requested):

WesleyGoncalves:
The PR changes worked for me.

shalvah:
> It actually worked for the first time I executed the script. Second time, it stops with this error
> 
> > rmdir(/path/to/project_dir/public/vendor/scribe/images): Directory not empty

That's an unrelated error; should have been fixed in recent versions.

shalvah:
> The PR changes worked for me.

Also, yes, the PR changes will work for you in the sense that Scribe won't crash, but _all_ your validation rules will be ignored. You can check and confirm that.

Closing this PR for the real fix.

PR #220, opened 2021-05-25
Fix link inside rst area (docs)

nanaya:
Missing letter aside, it should use rst syntax for linking, not markdown as it's inside an `eval_rst`.


shalvah:
Thanks. I'm pretty sure there's another instance of this too.

PR #225, opened 2021-05-29
update .gitattributes to ignore docs

lakuapik:
I think its safe to ignore the `docs` folder on `.gitattributes` file as it happends to save some space when installing scribe.

It will save approximately `1.4 MB` of disk space

image

Note: I also found that `pastel` also ignore the docs folder, see https://github.com/knuckleswtf/pastel/blob/476e86cd237c7256b5365efd844084a49075543a/.gitattributes#L1-L4

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,4 @@
 /phpunit.xml export-ignore
 /README.md export-ignore
 /body-params.png export-ignore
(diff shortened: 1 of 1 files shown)


shalvah:
Ah, thought I already did this. Thanks. Docs moving to a new repo in a few days anyway.

PR #235, opened 2021-06-08
Update UseApiResourceTags (fix invalid type error)

yaroslawww:
This comment:
```
* @apiResource 201 \App\Http\Resources\Advertisers\UserManagement\StaffResource
```
Throw error:
```
 Spatie\DataTransferObject\DataTransferObjectError 
 Invalid type: expected `Knuckles\Camel\Extraction\Response::status` to be of type `integer`, instead got value `201`, which is string..
```
So, to fix the added cast to integer status

### DIFF
--- a/src/Extracting/Strategies/Responses/UseApiResourceTags.php
+++ b/src/Extracting/Strategies/Responses/UseApiResourceTags.php
@@ -141,7 +141,7 @@ class UseApiResourceTags extends Strategy
     {
         $content = $tag->getContent();
         preg_match('/^(\d{3})?\s?([\s\S]*)$/', $content, $result);
-        $status = $result[1] ?: 0;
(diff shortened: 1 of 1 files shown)


shalvah:
Thanks. Will likely need to make the same changes for Transformers too. But what's with the I other changes files?

yaroslawww:
@shalvah 

In my case, only this problem gives an error. I am wary of making any changes to the transformers.

In index.blade.php, only the path to the assets files has been changed, since by default there is "../docs/" but in order not to publish the template and not change it, it is convenient to just use the relative path "./"

shalvah:
Alright, please remove the template changes from the PR, so I can merge. I don't really get your explanation, and it should be a separate PR, btw.

yaroslawww:
@shalvah  Done

PR #236, opened 2021-06-08
Set current relative path for html assets

yaroslawww:
If `output_path` set different value than `public/docs` - for example `public/docs/api` assets will not be loaded because aseets url will be https://my.site/docs/docs/<asset path>. After replace `../docs/` to simple `./` browse will loads correct relative path https://my.site/docs/api/<asset path> . Thus, there is no need to change the blade template.

### DIFF
--- a/resources/views/themes/default/index.blade.php
+++ b/resources/views/themes/default/index.blade.php
@@ -11,9 +11,9 @@
 
     <link href="https://fonts.googleapis.com/css?family=PT+Sans&display=swap" rel="stylesheet">
 
-    <link rel="stylesheet" href="../docs/css/theme-default.style.css" media="screen">
-    <link rel="stylesheet" href="../docs/css/theme-default.print.css" media="print">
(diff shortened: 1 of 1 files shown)


shalvah:
The reason for the ../docs path is for the static type docs to also work when the Laravel app is running. Merging this will break that, so no.

I generally don't recommend using a different path for the static docs output. Why are you using a custom path?

shalvah:
Ah, it's a tricky one. I'd prefer if people stuck with the default, but we shouldn't punish them for not doing so. And, yeah, you could publish the templates, but for a small change like this, you shouldn't have to.

How about adding a check for the output path? If it's the default, we can use `../docs`, otherwise go the other way. You could do this check in the HtmlWriter, then pass the `$assetPath` to the Blade template.

Btw, are you on v3? How did the migration go?

yaroslawww:
Version 3 work great for me. there were almost no problems with migration. Thank you for your work supporting the package!

yaroslawww:
@shalvah  It works great. Thank you very much!

PR #244, opened 2021-06-16
Update the doc url with in config.php

chrisliuqq:
I found the url in comment for `example_languages` in `config.php` is wrong, so I just modify it to the right one.
The url should be `https://scribe.knuckles.wtf/laravel/advanced/example-requests`  in `example_languages` section.


shalvah:
Thank you.

Issue #240, opened 2021-06-11
Try it out always returns error


koushik355:
Hi, thank you for such a fantastic library.
However, I'm currently facing an issue with the try it out button (send request one).
Please check the gif at the bottom. Even after filling the form, still, it says that validation error. 
Here is some information 

- PHP 8 
- Laravel 8
- Scribe version - 3.0.1
- Using form request to handling validation
- The doc for body params are generated automatically and I didn't define any `@bodyParam` in the doc.

![tryout](https://user-images.githubusercontent.com/66407490/121721134-8a433c00-cb01-11eb-9c3f-12a79e405fa2.gif)

shalvah:
Try checking what request is actually sent (Dev Tools' Network tab)?  What's in the body?

koushik355:
I did. Please check the attached video. The email is being in the formdata too. 

Uploading bandicam 2021-06-12 14-41-26-339.mp4…

koushik355:
I did. Please check the attached video. It does send the email in the body but still error. 


https://user-images.githubusercontent.com/66407490/121771505-59a8e400-cb8d-11eb-9ea6-7628224a0dd4.mp4

shalvah:
Looks like an encoding issue. I see the example request has a content type of JSON, but the actual request here uses form data (even though it sends a content-type of JSON). Looks like there's a bug that makes it send FormData.

Checking the code in tryitout.js, the deciding factor seems to be:

```js
    if (form.dataset.hasfiles === "0") {
        body = {};
        setter = (name, value) => _.set(body, name, value);
    } else {
        body = new FormData();
        setter = (name, value) => body.append(name, value);
    }
```

Try Inspect Element. What is the `data-hasfiles` attribute on the `<form>` element  wrapping that endpoint?

shalvah:
Thanks to @urbsny , this has been fixed in #245 . Will tag a new release later today.

koushik355:
Thank you very much @shalvah

shalvah:
3.0.3

PR #245, opened 2021-06-18
Fix body params format

urbsny:
Hi, thank you for such a fantastic library.
I encountered #240 issue.

`tryitout-3.0.2.js` is checking hasFiles attribute is "0".
```javascript
    if (form.dataset.hasfiles === "0") {
        body = {};
        setter = (name, value) => _.set(body, name, value);
    } else {
        body = new FormData();
        setter = (name, value) => body.append(name, value);
    }
```

But generated html has not "0" or "1"
```html
<form id="form-POSTapi-v1-login" data-method="POST" data-path="api/v1/login" data-authed="0" data-hasfiles="" data-headers="{&quot;Content-Type&quot;:&quot;application\/json&quot;,&quot;Accept&quot;:&quot;application\/json&quot;}" onsubmit="event.preventDefault(); executeTryOut('POSTapi-v1-login', this);">
```


shalvah:
Ah, thank you.

PR #249, opened 2021-06-22
Fix the boundary is missing in the headers.

### LINKED ISSUE #232: Can file input be tested on the docs?
- [x] I've read the [documentation](https://scribe.rtfd.io) and I can't find details on how to achieve this.

I am working on a project, and an API's request has a parameter for type file. Can that be possible with scribe?

As, I am testing it I am getting an error:
scribe-bug

I understand that this is a warning message from the API saying that the Boundary is missing in the headers
Error: Missing boundary in multipart/form-data POST data

So, as the request of the API has the File input, it changes the content type to Form Data, but it is not sending the boundary with it.

Can this be fixed?

urbsny:
Hi! I fixed #232.
When I sent FormData without setting the Content-type, the Content-type was automatically set with boundary.
But I'm not sure it's best way to fix. So If you don't like it, feel free to ignore it.


shalvah:
Makes sense, thank you.👍

Issue #260, opened 2021-07-01
TDD or Test-Driven Documentation


ajcastro:
Hi, I just like to share this idea.

This is a documentation principle where you generate your api docs by running phpunit tests instead of docblocks annotations in the controller. 

Benefits
- Updating your tests updates your docs automatically.
- Better workflow, instead of writing docblock annotations and parameters in controller, you can auto-generate documentation from the tests performed. 
- Less comments or visual debt in controller codebase. (Some annotations are still needed but annotations can be put in the test classes instead)
- Follows the principle "If it is not tested, it does not exist." making sure your docs and test are in sync.

Currently, there is a package that has this principle/workflow, called Enlighten, https://github.com/StydeNet/enlighten/. 
But scribe output looks better and easier to navigate.
Scribe output has more fun features like the "Try it out".
Scribe has support for OpenAPI and Postman collection.

So I am wondering if scribe can adopt test-driven documentation so I can drop Enlighten.

shalvah:
That's an interesting idea. I don't see myself exploring it anytime soon because I don't have the bandwidth right now, but you can always send in a proof of concept.

Theoretically, you could add it to the existing Scribe architecture as a new strategy, say `GetDetailsFromTests`. A limitation of this is that it will be run once for each endpoint, so you have to only run the tests for that endpoint.

A second limitation is that the strategy has to be registered with a specific stage (headers, query parameters etc), but it's easy to get around that by mutating the `$endpointData` object directly.

ajcastro:
Hi @shalvah , thanks that you like the idea! 
Maybe I can ask the author of Enlighten if he is interested to collaborate and help us implement the feature he has done in Enlighten. I hope he would help us for the sake of the open-source community. If ever, I am also interested to help to implement that feature.

ajcastro:
Hi, @shalvah I plan to work on this, do you have a public repository that uses scribe? So I can start to familiarize myself and deep dive into how it works. It would be better if the repo is using most of the features of scribe. Although I also created a sample laravel app and installed scribe but it would be better if I can look on others work.

ajcastro:
I found it, haha. https://github.com/knuckleswtf/TheSideProjectAPI

shalvah:
Yeah, it's slightly out-of-date, but I'll be updating it in a few hours.

ajcastro:
Scribe-TDD is ready for testing. I was hoping we can get some beta testers. https://github.com/ajcastro/scribe-tdd

[...]

PR #263, opened 2021-07-02
Separate the concern of getting groupedEndpoints by creating different strategies

ajcastro:
Separate the concern of getting groupedEndpoints by creating different strategies for getting groupedEndpoints. 

Before, the command class `GenerateDocumentation` contains all the helper methods in getting the groupedEndpoints.
In this PR, I refactor it by creating different strategies in getting the groupedEndpoints. 

When I read the code, currently there are two ways of getting the groupedEndpoints:
1. First is extracting the endpoints from the Laravel app, I call this strategy class `GroupedEndpointsFromApp`
2. Then the other one is getting it from the generated YAML from the camelDir, I call this `GroupedEndpointsFromCamelDir`. 

So first there is `GroupedEndpointsFactory` that will return the appropriate strategy, either `GroupedEndpointsFromApp` if it `isForcing` or `shouldExtract` is true. Otherwise, it will use `GroupedEndpointsFromCamelDir`. 

This PR is important because later on we can add another strategy on getting the groupedEndpoints which is I plan to do next, 
`GroupedEndpointsFromTests`, which returns the groupedEndpoints from the performed phpunit tests which I mention in TDD or Test-Driven Documentation, https://github.com/knuckleswtf/scribe/issues/260

This PR seems to have a lot of file changes but basically what I just did here is just moved the helper methods together for the `GroupedEndpointsFromApp`.  Then I created another strategy for `GroupedEndpointsFromCamelDir` and create `GroupedEndpointsFactory`

### DIFF
--- a/src/Commands/GenerateDocumentation.php
+++ b/src/Commands/GenerateDocumentation.php
@@ -6,23 +6,19 @@ use Illuminate\Console\Command;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Str;
+use Knuckles\Camel\Camel;
 use Knuckles\Camel\Extraction\ExtractedEndpointData;
 use Knuckles\Camel\Output\OutputEndpointData;
-use Knuckles\Camel\Camel;
-use Knuckles\Scribe\Extracting\Extractor;
 use Knuckles\Scribe\Extracting\ApiDetails;
-use Knuckles\Scribe\Matching\MatchedRoute;
+use Knuckles\Scribe\Extracting\Extractor;
+use Knuckles\Scribe\GroupedEndpoints\GroupedEndpointsFactory;
 use Knuckles\Scribe\Matching\RouteMatcherInterface;
 use Knuckles\Scribe\Tools\ConsoleOutputUtils as c;
 use Knuckles\Scribe\Tools\DocumentationConfig;
 use Knuckles\Scribe\Tools\ErrorHandlingUtils as e;
 use Knuckles\Scribe\Tools\Globals;
 use Knuckles\Scribe\Tools\Utils;
-use Knuckles\Scribe\Tools\Utils as u;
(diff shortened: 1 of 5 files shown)


shalvah:
Well, that was fast.

shalvah:
Hmmm, makes sense. I'll need to do a bit of testing though. Sadly there aren't enough (or any) automated tests around this part of the code yet.

ajcastro:
@shalvah Sure. Thanks! I'm excited about this. 😄

ajcastro:
@shalvah I created another refactoring, https://github.com/ajcastro/scribe/pull/1

shalvah:
I've added some more tests, so I'm taking a look at this now, and it's good.👍

However, I'm not so sure about the full-fledged tests approach yet. Two major reasons:
1. I'm not sure if I'm ready to maintain this. In fact, considering my plans and goals right now, I don't think I can. (And yes, you could say _you would_, but I won't be able to not worry about what's in my codebase and how everything interacts.)
2. I haven't decided how it fits into Scribe  For example, I need to decide if this is the primary approach we'll point people to. Multiple places in the docs may need to be updated to cater for this, and I need to decide how. And then there'll be people that will have issues and questions. I'm not yet convinced of the tests strategy, so I won't be very helpful there.

On the other hand, thanks to the whole Camel thing + the factory approach you just added, it's easier to do this in userland. Since the GroupedEndpoints factory is injected into the `handle()` method, you can bind a new factory in the DI container that includes the tests strategy instead, as long as output is in the Camel format. Hell, you could probably just pull in most of the code from the other package and convert the output into Camel objects.

So here's what I recommend—let's make it a plugin. You can get this to work and publish it, and in the docs, I'll point people to it if they prefer that approach. Over time, if it does well, and I take a liking to it, we can then bundle it with Scribe out of the box. 👍

ajcastro:
Thanks @shalvah . I like the idea of making it a plugin. I'll let you know if it is ready, currently, I am still working on it.

PR #272, opened 2021-07-09
Add getDocBlocks() in RouteDocBlocker

ajcastro:
This PR adds a method `getDocBlocks(Route $route, $className, $methodName = null)` in `RouteDocBlocker`. 
The reason behind this so that we can pass a different `className` and `methodName`. In my plugin, [scribe-tdd](https://github.com/ajcastro/scribe-tdd), The docblocks annotations are used in the test classes and methods. 

Therefore, I plan to create a strategy with implementation to use this like so: 
```php
RouteDocBlocker::getDocBlocks($route, 'Tests\Features\Http\UserControllerTest', 'test_users_index');
```
This will parse the docblocks in the test class and its method.

You can check a fork of the `TheSideProjectAPI` where I move the docblock annotations to the test classes:
https://github.com/ajcastro/TheSideProjectAPI/pull/1

### DIFF
--- a/src/Extracting/RouteDocBlocker.php
+++ b/src/Extracting/RouteDocBlocker.php
@@ -38,7 +38,6 @@ class RouteDocBlocker
             [$className, $methodName] = $className;
         }
 
-        [$className, $methodName] = u::getRouteClassAndMethodNames($route);
         $normalizedClassName = static::normalizeClassName($className);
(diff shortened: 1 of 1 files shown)


ajcastro:
Thanks @shalvah I have used this well. Working strategy for metadata, https://github.com/ajcastro/scribe-tdd/blob/master/src/Extracting/Strategies/Metadata/GetFromDocBlocksFromScribeTdd.php

Issue #279, opened 2021-07-17
Call a dynamic model


ddaattttaa:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**
1. In my controller, i call the model like that 
```
/**
 * @group Brand
 * Class BrandController
 * @package App\Http\Controllers
 */
class BrandController extends BaseController
{

    protected $model = Brand::class;
    protected $collectionResource = BrandResourceCollection::class;


    /**
     * Display a listing of the resource.
     *
     * @param Request $request
     * @return JsonResponse
     */
    public function index(Request $request): JsonResponse
    {
        $dataBuilder = $this->model::query();
            return $this->successResponse(new $this->collectionResource($dataBuilder->paginate()), "$this->model list");
    }
```
2. Then I ran `php artisan scribe:generate` i have an error and when i see the file **Knuckles\Scribe\Extracting\Strategies\GetFromInlineValidatorBase** at line 54 the variable **$rvalue** split **$this** and **model** in the object **PhpParser\Node\Expr\StaticCall**
**$rvalue = $validationAssignmentExpression->expr;**

**Screenshots and stack traces:**
```
ⓘ Processing route: [GET] api/v1/brands

   TypeError

  end(): Argument #1 ($array) must be of type array, null given

  at vendor/knuckleswtf/scribe/src/Extracting/Strategies/GetFromInlineValidatorBase.php:54
     50▕                 $validationRules = $rvalue->args[0]->value;
     51▕                 break;
     52▕             } else if (
     53▕                 // Try $validator = Validator::make(...)
  ➜  54▕                 $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts) == "Validator"
     55▕                 && $rvalue->name->name == "make"
     56▕             ) {
     57▕                 $validationRules = $rvalue->args[1]->value;
     58▕                 break;

```

**My environment:**
 - PHP version (from `php -v`): v8.0.7
 - Framework (Laravel/Lumen): Laravel
 - Laravel/Lumen version (from `composer show laravel/framework` or `composer show laravel/lumen-framework`): v8.50.0
 - Scribe version (from `composer show knuckleswtf/scribe`): 3.6.0

have you an idea ?
Thanks

ddaattttaa:
Hello, 

Thank you for you quickly fix !

PR #281, opened 2021-07-20
fix https://github.com/knuckleswtf/scribe/issues/279

tommica:
I also encountered this [bug](https://github.com/knuckleswtf/scribe/issues/279), and this solves the issue.

### DIFF
--- a/src/Extracting/Strategies/GetFromInlineValidatorBase.php
+++ b/src/Extracting/Strategies/GetFromInlineValidatorBase.php
@@ -51,7 +51,7 @@ class GetFromInlineValidatorBase extends Strategy
                 break;
             } else if (
                 // Try $validator = Validator::make(...)
-                $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts) == "Validator"
+                $rvalue instanceof Node\Expr\StaticCall && $rvalue->class->parts && end($rvalue->class->parts) == "Validator"
                 && $rvalue->name->name == "make"
             ) {
                 $validationRules = $rvalue->args[1]->value;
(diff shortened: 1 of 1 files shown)


src/Extracting/Strategies/GetFromInlineValidatorBase.php
```diff
@@ -51,7 +51,7 @@ public function lookForInlineValidationRules(ClassMethod $methodAst): array
                 break;
             } else if (
                 // Try $validator = Validator::make(...)
-                $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts) == "Validator"
+                $rvalue instanceof Node\Expr\StaticCall && $rvalue->class->parts && end($rvalue->class->parts) == "Validator"
```
shalvah, line 54: I think this is better

```suggestion
                $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts ?: []) == "Validator"
```

src/Extracting/Strategies/GetFromInlineValidatorBase.php
```diff
@@ -51,7 +51,7 @@ public function lookForInlineValidationRules(ClassMethod $methodAst): array
                 break;
             } else if (
                 // Try $validator = Validator::make(...)
-                $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts) == "Validator"
+                $rvalue instanceof Node\Expr\StaticCall && end($rvalue->class->parts ?: []) == "Validator"
```
shalvah, line 54: Ah no, yours was better.

```suggestion
                $rvalue instanceof Node\Expr\StaticCall && $rvalue->class->parts && end($rvalue->class->parts) == "Validator"
```

PR #282, opened 2021-07-21
Support for spatie/data-transfer-object 3

### LINKED ISSUE #241: Compatible with spatie/dto ^3
Hello! Is it possible to implement compatibility with the ^3 version?

```
  Problem 1
    - knuckleswtf/scribe[3.0.0, ..., 3.0.2] require spatie/data-transfer-object ^2.6 -> found spatie/data-transfer-object[2.6.0, ..., v2.x-dev] but it conflicts with your root composer.json require (^3.3).
    - Root composer.json requires knuckleswtf/scribe ^3.0 -> satisfiable by knuckleswtf/scribe[3.0.0, 3.0.1, 3.0.2].

```

punyflash:
As `spatie/data-transfer-object` dropped support for `DataTransferCollection`, which was used to iterate `ResponseCollection` I decided to reimplement iterator functionality dirrectly in `BaseDTOCollection`. I still not sure this is the best sollution as now `BaseDTOCollection` looks a bit messy, but this will resolve #241.

There was issues with constructor for `BaseDTOCollection` on `dto@3`. On some tests given collection was an instance of `ResponseCollection` or null. So the type checks in 32 and 33 lines of `BaseDTOCollection` are necessary.

Tested with `dto@2` and `dto@3` - all tests pass successfully.

### DIFF
--- a/camel/BaseDTOCollection.php
+++ b/camel/BaseDTOCollection.php
@@ -2,22 +2,36 @@
 
 namespace Knuckles\Camel;
 
+use ArrayAccess;
 use ArrayIterator;
+use Countable;
 use Illuminate\Support\Arr;
-use Spatie\DataTransferObject\DataTransferObjectCollection;
+use Iterator;
+use Spatie\DataTransferObject\DataTransferObject;
 
 /**
  * @template T of \Spatie\DataTransferObject\DataTransferObject
  */
-class BaseDTOCollection extends DataTransferObjectCollection
+abstract class BaseDTOCollection implements
+    ArrayAccess,
+    Iterator,
+    Countable
 {
+    protected ArrayIterator $iterator;
+
     /**
      * @var string The name of the base DTO class.
      */
     public static string $base = '';
 
-    public function __construct(array $collection = [])
+    /**
+     * @param BaseDTOCollection|array $collection
+     */
+    public function __construct($collection = [])
     {
+        if (is_null($collection)) $collection = [];
+        if (!is_array($collection)) $collection = $collection->toArray();
+
         // Manually cast nested arrays
         $collection = array_map(
             function ($item) {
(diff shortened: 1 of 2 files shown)


shalvah:
That's an interesting approach, but how about we just pull in illuminate/support and use an Illuminate collection instead? I don't really like us having all this boilerplate in the BaseDTOCollection class?

punyflash:
> That's an interesting approach, but how about we just pull in illuminate/support and use an Illuminate collection instead? I don't really like us having all this boilerplate in the BaseDTOCollection class?

I do believe that thats possible to refractor using Illuminate collections, but this requires much more refactoring. I will not take the risk doing that, as I'm not really familiar with library insides. My approach was kind of straightforward - to resolve issue with the least of intervention.

shalvah:
Actually, I think it would be *easier*. When you look at the `BaseDTOCollection`, we only have three pieces of custom functionality:
- `__construct`, where we make sure to cast each item first
- `concat` (same casting thing)
- `sortBy`

Illuminate collections already come with a sortBy, so that would replace the last piece. Essentially, the new `BaseDTOCollection` would have an inner collection instance, and contain just the first two methods plus a `__call` to forward method calls to the inner collection. Additionally, the code in `ResponseCollection` would be simpler as well.

I understand you might be reluctant to make changes, but this PR adds nearly 100 lines of code to the class, extra boilerplate that I don't think is needed, and I'd rather not have in my way when I'm trying to figure out the actual logic of the code. Using a collection avoids that.

shalvah:
Thanks for the work! I ended up implementing what I wanted in #285. It _was_ easier, although that's partly because I'm familiar with my codebase. Also gave me the opportunity to fix a few things.

Tests pass with dto v3 (via GH Actions), but I can't guarantee that the collection is the only change needed, so I'm taking your word for it!

PR #287, opened 2021-07-23
Fix https://github.com/knuckleswtf/scribe/pull/281 (again)

tommica:
I encountered another variant of this bug, where the "parts" might not always be defined, causing a fatal error. This solution fixes it.

### DIFF
--- a/src/Extracting/Strategies/GetFromInlineValidatorBase.php
+++ b/src/Extracting/Strategies/GetFromInlineValidatorBase.php
@@ -51,7 +51,7 @@ class GetFromInlineValidatorBase extends Strategy
                 break;
             } else if (
                 // Try $validator = Validator::make(...)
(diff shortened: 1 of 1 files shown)


src/Extracting/Strategies/GetFromInlineValidatorBase.php
```diff
@@ -51,7 +51,7 @@ public function lookForInlineValidationRules(ClassMethod $methodAst): array
                 break;
             } else if (
                 // Try $validator = Validator::make(...)
-                $rvalue instanceof Node\Expr\StaticCall && $rvalue->class->parts && end($rvalue->class->parts) == "Validator"
+                $rvalue instanceof Node\Expr\StaticCall && isset($rvalue->class->parts) && $rvalue->class->parts && end($rvalue->class->parts) == "Validator"
```
shalvah, line 54: empty = isset + empty

```suggestion
                $rvalue instanceof Node\Expr\StaticCall && !empty($rvalue->class->parts) && end($rvalue->class->parts) == "Validator"
```

PR #288, opened 2021-07-23
Fix factory error detection

SquareBeard:
Fixes a problem caused by a backslash escaping the next character in a string.
The bug prevented the exception on line 204 to be thrown.

### DIFF
--- a/src/Tools/Utils.php
+++ b/src/Tools/Utils.php
@@ -200,7 +200,7 @@ class Utils
             try {
                 $factory = factory($modelName);
             } catch (\Throwable $e) {
-                if (Str::contains($e->getMessage(), "Call to undefined function Knuckles\Scribe\Tools\\factory()")) {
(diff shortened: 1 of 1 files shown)


src/Tools/Utils.php
```diff
@@ -200,7 +200,7 @@ public static function getModelFactory(string $modelName, array $states = [], ar
             try {
                 $factory = factory($modelName);
             } catch (\Throwable $e) {
-                if (Str::contains($e->getMessage(), "Call to undefined function Knuckles\Scribe\Tools\factory()")) {
+                if (Str::contains($e->getMessage(), "Call to undefined function Knuckles\Scribe\Tools\\factory()")) {
```
shalvah, line 203: I think all the \ should be escaped, not just the last.

shalvah (changes requested):

PR #297, opened 2021-08-18
fix for validating digits_between rule

piotranon:
### Hi i found out that generating docs with `digits_between:min,max` rule is not working right

Example rules from `Request:Class`
```php
class CopyProductRequest extends FormRequest
{
    /**
     * Determine if the user is authorized to make this request.
     *
     * @return bool
     */
    public function authorize()
    {
        return true;
    }
    /**
     * Get the validation rules that apply to the request.
     *
     * @return array
     */
    public function rules()
    {
        return [
              'gtin' => 'nullable|digits_between:0,20',
        ];
    }
}
```
Is throwing a error:
```
Failed processing route: [POST] api/v1/projects/{project}/products - Exception encountered.

   Knuckles\Scribe\Exceptions\ProblemParsingValidationRules 

  Problem processing validation rules for the param `gtin`

  at D:\app\vendor\knuckleswtf\scribe\src\Exceptions\ProblemParsingValidationRules.php:11
      7▕ class ProblemParsingValidationRules extends \RuntimeException implements ScribeException
      8▕ {
      9▕     public static function forParam(string $paramName,  Throwable $innerException): ProblemParsingValidationRules
     10▕     {
  ➜  11▕         return new self(
     12▕             "Problem processing validation rules for the param `$paramName`",
     13▕             0, $innerException);
     14▕     }
     15▕ }

  1   D:\app\vendor\knuckleswtf\scribe\src\Extracting\ParsesValidationRules.php:107
      Knuckles\Scribe\Exceptions\ProblemParsingValidationRules::forParam("gtin", Object(InvalidArgumentException))

  2   D:\app\vendor\fakerphp\faker\src\Faker\Provider\Base.php:86
      InvalidArgumentException::("randomNumber() can only generate numbers up to mt_getrandmax()")
```

Faker `numberBetween` method is already creating random number in specified range.
Faker `randomNumber` is generating a number but first arg is length of generated number.

So if first you generate number between 0 and 20 you can get 20, and `randomNumber` method will try to generate random number with length of 20.  
( and that is throwing this error because `randomNumber` cannot generate number bigger than `mt_getrandmax() = 2147483647` )

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -344,7 +344,7 @@ trait ParsesValidationRules
                     break;
                 case 'digits_between':
                     $parameterData['description'] .= ' ' . $this->getDescription($rule, [':min' => $arguments[0], ':max' => $arguments[1]]);
-                    $parameterData['setter'] = fn() => $this->getFaker()->randomNumber($this->getFaker()->numberBetween(...$arguments), true);
+                    $parameterData['setter'] = fn() => $this->getFaker()->numberBetween(...$arguments);
                     $parameterData['type'] = 'number';
                     break;
 


shalvah:
Thank you for volunteering the fix, but I'm thinking there's a misunderstanding. 🤔 Isn't the `digits_between` rule about the length of the number, not the value? [The doc](https://laravel.com/docs/8.x/validation#rule-digits-between) says:

> The field under validation must be numeric and must have a length between the given min and max.

If you want a number between 0-20, shouldn't you be using [the `between` rule](https://laravel.com/docs/8.x/validation#rule-between)? Probably as `integer|between:0,20`.

shalvah:
Yeah, I think this is correct as it is. You'll notice that some of the tests are failing on that rule (the generated example is validated by Laravel). You probably want `between`.

Thanks for the effort, though!

piotranon:
What if you want to use `digits_between` because you need to validate that input number is with length of max 20 digits?
Is there a way to pass a validation rule?

piotranon:
> Yeah, I think this is correct as it is. You'll notice that some of the tests are failing on that rule (the generated example is validated by Laravel). You probably want `between`.
> 
> Thanks for the effort, though!

What if you want to use `digits_between` because you need to validate that input number is with length of max 20 digits?
Is there a way to pass a validation rule?

Because randomNumber is generating number only to max 10 digits.

shalvah:
Yeah, that makes sense, since mt_getrandmax() is 10 digits. If you want more, the type should be a string, since PHP can't handle larger numbers. And the value should be generated using [numerify](https://fakerphp.github.io/formatters/numbers-and-strings/), so something like:

```php
$randomLength = rand($min, $max);
$value = $faker->numerify(str_repeat("#", $randomLength));
```

piotranon:
> ```
> numerify
> ```

so will you add a fix for that?

shalvah:
Perhaps. It's a relatively small change, but we'll also need to change `$parameterData['type']` to `"string"`. We'll also need to make the same change for the `digits` rule (only difference is that the length is not random).

I'll get to it when I can, but you can also open a PR.

PR #311, opened 2021-09-02
Minify Resource Assets

ezralazuardy:
### Detail

- Minify the `css` and `js` resource asset files using [minify](https://www.npmjs.com/package/minify) for faster load
- Added link preload tag for assets preloading
- Fixed some `css` redundant style

### Test

`composer test` result:

image


shalvah:
Well, I appreciate the effort, but I'm not sure about the modification part. Right now, I'm not ready to introduce another build tool into the whole thing, and when I have to debug, I'd rather not have to fiddle with minified code or source maps. Can't tell you enough how I hate that the JS is minified.

I do like the preload and general formatting parts. So maybe you could remove the modification part, and I can merge this.

Also (if you have the time), could you help strip the lodash? I linked to the full Lodash build because I didn't want to figure out how to make a custom build, but we really only use a couple of functions *in fact, I think it's only `_.set`). If you could find a CDN build of just that, it would be great; otherwise, we could make a custom build and include it here. If you can't do either, no problem.

ezralazuardy:
sure, i'll make other pr for the lodash

PR #312, opened 2021-09-02
Group radio buttons when setting query parameters

stelar7:
Radio buttons would call `_.set` for each of them, meaning that the value of the parameter would always be the set to the same as the last added. (`false` in the case of `boolean` validations)

With this change, the selected value is set instead.

### DIFF
--- a/resources/js/tryitout.js
+++ b/resources/js/tryitout.js
@@ -178,10 +178,12 @@ async function executeTryOut(endpointId, form) {
     const query = {};
     const queryParameters = form.querySelectorAll('input[data-component=query]');
     queryParameters.forEach(el => _.set(query, el.name, el.value));
+    
+    // Group radio buttons by their name, and then set the checked value from that group
     Array.from(queryParameters)
         .filter(el => el.type === "radio")
         .reduce(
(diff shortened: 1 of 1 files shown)


shalvah (changes requested):
I like how clean this code looks, but can you add a comment or two that explains what's going on at a high-level?

Also, if we're _undoing_ something, isn't it possible to avoid it in the first place (ie avoid setting the radio button)?

shalvah (approved):

resources/js/tryitout.js
```diff
@@ -178,6 +178,21 @@ async function executeTryOut(endpointId, form) {
     const query = {};
     const queryParameters = form.querySelectorAll('input[data-component=query]');
     queryParameters.forEach(el => _.set(query, el.name, el.value));
```
shalvah, line 180: You know, I think this would have just been simpler:

```js
    queryParameters.forEach(el => {
        if (el.type !== 'radio' || (el.type === 'radio && el.checked)) {
            _.set(query, el.name, el.value));
        }
    });
```

I'll change it. Simpler is better. 😅

PR #324, opened 2021-09-15
Feature/field examples. Closes #322

### LINKED ISSUE #322: Should the example params/body be used in the try-it-out fields?
- [x] I've read the [documentation](https://scribe.knuckles.wtf/laravel) and I can't find details on how to achieve this.

I was wondering why the fields in the try-it-out aren't using the same values that are used in the example requests. It is actually really to add these examples into the field by just passing the example field to `scribe::components.field-details` and setting it as the `value` for the corresponding inputs.

You can see I did that as a test checking if it worked:

Screen Recording 2021-09-14 at 09 25 33 AM

Thoughts on this?

james2doyle:
Closes #322

On top of the examples being put in, I added a test for the field-details component


james2doyle:
hmm weird, not sure why a class is just completely missing in lumen

shalvah:
`InteractsWithViews` - Lumen doesn't come with views support by default. Also, was that class present in Laravel 6? That's where the test fails.

james2doyle:
Should I just remove the test then? Seems like its only applicable to half to app

shalvah:
Yeah, I  think we can remove the test/

Issue #330, opened 2021-09-25
PHP Content-Type header not displayed in response examples.


rickdubiel:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

I can't seem to get the PHP Content-Type header to display in any of the response examples. I've also noticed that it's not being displayed in the demo as well for ONLY the PHP language.

https://demo.scribe.knuckles.wtf/

For example it show's up in bash (and javascript, python, and ruby):

![image](https://user-images.githubusercontent.com/18044230/134751180-09de9692-f0fe-495e-a20a-5c231906feee.png)

But not in php:

![image](https://user-images.githubusercontent.com/18044230/134751205-6304a8f0-737c-43b4-90a6-37ddcec53cd4.png)

Thanks!

shalvah:
- Those are request examples, not response examples.
- The content type isn't included, because the Guzzle library, which the PHP examples use, automatically sets it based on the data you provide. For example, passing in `json => [...]` will set it to `application/json`.
- The screenshot you shared is of a GET request, so there's no body **content**;there and hence no content-type, but if you check some other type, you'll see what I mean.
- If you don't agree with this and will prefer to include your headers manually, you can override the language templates with yours. Can't get you the link rn, but it's described in the docs, under Advanced Customisation, I think.

rickdubiel:
> Those are request examples, not response examples.
My bad, I did mean request examples..

> The content type isn't included, because the Guzzle library, which the PHP examples use, automatically sets it based on the data you provide. 
That's interesting, but wouldn't you think we'd still want to display the Content-Type to whoever is integrating the api? Especially for folks like me who wasn't aware that Guzzle did that behind the scenes? I don't know, just feel like it'd be more clear that way.

> The screenshot you shared is of a GET request, so there's no body content;there and hence no content-type, but if you check some other type, you'll see what I mean.
Sorry, bad examples but there is also a POST request on that demo without the Content-Type header as well (even though it's not shown due to bullet #2 above).

![image](https://user-images.githubusercontent.com/18044230/134752710-e5ffa196-1a7f-4426-9e19-3cc36ac3836e.png)

> If you don't agree with this and will prefer to include your headers manually, you can override the language templates with yours. Can't get you the link rn, but it's described in the docs, under Advanced Customisation, I think.
Thank you for pointing this out! This will have to be the route I take. 

Thank you so very much for the quick response! Love the package!

PR #331, opened 2021-09-25
Adjusted language examples to properly display application/x-www-form…

rickdubiel:
This PR provides support for application/x-www-form-urlencoded requests to be displayed properly within the various language examples. The app I'm working with is using Laravel Passport and I was needing a way to display to users how to obtain an access token via the following URL (which uses application/x-www-form-urlencoded under the hood).

https://laravel.com/docs/8.x/passport#retrieving-tokens

I'm not exactly proficient in all of the languages but used my best judgement. Below are some of the before/after screenshots. With the exception of python, application/x-www-form-urlencoded expects parameters instead of json so those are what the adjustments are focused on.

bash (before)
image

bash (after)
image

javascript (before)
image

javascript (after)
image

php (before)
image

php (after)
image

The php demo was hiding the Content-Type header, and I know you (@shalvah) mentioned here (https://github.com/knuckleswtf/scribe/issues/330) that this is because the Grizzly library that the examples are using automatically add those for you, but I still believe it'd be best to explicitly display those to the users integrating with the documentation. Anything to make it easier right? Plus, the Guzzle docs just mention that Content-Type headers are only set when no Content-Type header is already present. Well in this case one is being explicitly provided and should be displayed. Another part of the issue here, and maybe the bigger part of the issue, was that there was no way to display the body type of "form_params" instead of "json" (which the oauth/token example doesn't work when using json).

https://docs.guzzlephp.org/en/stable/request-options.html#form-params

python (before)
image

python (after)
image

Not a python user so I hope is correct.

Thanks for your time!

### DIFF
--- a/resources/views/partials/example-requests/bash.md.blade.php
+++ b/resources/views/partials/example-requests/bash.md.blade.php
@@ -23,7 +23,11 @@ curl --request {{$endpoint->httpMethods[0]}} \
 @endforeach
 @endforeach
 @elseif(count($endpoint->cleanBodyParameters))
+@if ($endpoint->headers['Content-Type'] == 'application/x-www-form-urlencoded')
+    --data "{!! http_build_query($endpoint->cleanBodyParameters, '', '&') !!}"
+@else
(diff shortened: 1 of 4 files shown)


shalvah:
Okay, this should be fine.

PR #333, opened 2021-09-28
Fixed: case where null/optional query strings are sent

james2doyle:
When optional query string parameters are present, they are sent as empty when a request is made. This is interpreted by laravel as these fields are set to `null`. This is an issue when it comes to grabbing a default value from a request.

For example, a request with an optional `order` will be sent as `/api?order=` when the field is empty in the form (current implementation) which means the code `$request->query('order', 'desc')` will return `null` and not `'desc'` because the field was intentionally sent as `null` in the query string.

Anyway, I thought this change might be useful as it just avoids sending `null` in query strings. It did help find a flaw in my code so maybe the argument can be made that it can stay...


shalvah:
I want to merge this, but I can't reproduce the problem in the first place. Optional unset parameters are not sent in my case.

james2doyle:
Oh strange. Using the latest Chrome?

shalvah:
Yes. Gimme an example I can replicate.

shalvah:
Closing this...if you've got a counter case, you can reopen.

PR #364, opened 2021-11-11
Feat nested relations

TomaszKisiel:
# Overview

Allows for factoring nested relations with dot notation. Supports `HasMany` and `BelongsToMany` relations type. In case of custom data in pivot table you should create helper function in your factory which must be names as `pivot{name of relationship method}`. 

### Example use case:

`employees` <-- one to many --> `places` <-- many to many --> `tasks`.

Each `employee` has assigned many `places` of work and each `places` belongs to many `tasks`. Pivot table between `places` and `tasks` has additional column with `salary`. Resource returns those relations with information about salary from pivot.

# Models

### Employee
```php
class Employee extends Model
{
    use HasFactory;

    public function places(): HasMany
    {
        return $this->hasMany(Place::class);
    }
}
```

### Place
```php
class Place extends Model
{
    use HasFactory;

    public function tasks(): BelongsToMany
    {
        return $this->belongsToMany(Task::class);
    }
}
```

### Task
```php
class Task extends Model
{
    use HasFactory;

    public function places(): BelongsToMany
    {
        return $this->belongsToMany(Place::class);
    }
}
```

# Factories

### EmployeeFactory
```php
class EmployeeFactory extends Factory
{
    protected $model = Employee::class;

    public function definition(): array
    {
        return [
           'name' => $this->faker->name(),
        ];
    }
}
```

### PlaceFactory
```php
class PlaceFactory extends Factory
{
    protected $model = Place::class;

    public function definition(): array
    {
        return [
           'location' => $this->faker->word(),
        ];
    }

    public function pivotTasks(): array
    {
        return [
            'salary' => $this->faker->randomNumber(),
        ];
    }
}
```

### TaskFactory
```php
class TaskFactory extends Factory
{
    protected $model = Task::class;

    public function definition(): array
    {
        return [
           'description' => $this->faker->word(),
        ];
    }

    public function pivotPlaces(): array
    {
        return [
            'salary' => $this->faker->randomNumber(),
        ];
    }
}
```

# Resources

### EmployeeResource
```php
class EmployeeResource extends JsonResource
{
    public function toArray($request): array
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
            'places' => PlaceResource::collection($this->places)
        ];
    }
}
```

### PlaceResource
```php
class PlaceResource extends JsonResource
{
    public function toArray($request): array
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
            'tasks' => $this->tasks->map(function (Task $task) {
                return [
                    'task_id' => $task->id,
                    'salary' => $task->pivot->salary
                ];
            })
        ];
    }
}
```

# Controllers

### EmployeeController
```php
/**
 * @group Employee
 * @authenticated
 */
class EmployeeController extends Controller
{
    /**
     * Display a listing of the employees.
     *
     * @apiResourceCollection App\Http\Resources\EmployeeResource
     * @apiResourceModel App\Models\Escort paginate=10 with=places.tasks
     *
     * @param Request $request
     * @return AnonymousResourceCollection
     */
    public function index(Request $request): AnonymousResourceCollection
    {
        $perPage = $request->input('per-page', 10);
        $employees = Employee::paginate($perPage);

        return EmployeeResource::collection($employees);
    }
}
```

# Result

The result of current setup will be as follow:

```json
{
    "data": [
        {
            "id": 1,
            "name": "Adam",
            "places": [
                {
                    "location": "Warszawa",
                    "tasks": [
                        {
                            "task_id": 1,
                            "salary": 128
                        }
                    ]
                }
            ]
        },
        {
            "id": 2,
            "name": "Romek",
            "places": [
                {
                    "location": "Kraków",
                    "tasks": [
                        {
                            "task_id": 2,
                            "salary": 48392
                        }
                    ]
                }
            ]
        }
    ],
    "links": {
            
    },
    "meta": {
    
    }
}
```

### DIFF
--- a/src/Tools/Utils.php
+++ b/src/Tools/Utils.php
@@ -2,7 +2,6 @@
 
 namespace Knuckles\Scribe\Tools;
 
-use _PHPStan_76800bfb5\Nette\PhpGenerator\PhpFile;
 use Closure;
 use DirectoryIterator;
 use Exception;
--- a/src/Tools/Utils.php
+++ b/src/Tools/Utils.php
@@ -2,10 +2,13 @@
 
 namespace Knuckles\Scribe\Tools;
 
+use _PHPStan_76800bfb5\Nette\PhpGenerator\PhpFile;
 use Closure;
 use DirectoryIterator;
 use Exception;
 use FastRoute\RouteParser\Std;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Routing\Route;
 use Illuminate\Support\Str;
 use Knuckles\Scribe\Exceptions\CouldntFindFactory;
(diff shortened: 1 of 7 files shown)


src/Tools/Utils.php
```diff
@@ -190,13 +192,31 @@ public static function getModelFactory(string $modelName, array $states = [], ar
             /** @var \Illuminate\Database\Eloquent\Factories\Factory $factory */
             $factory = call_user_func_array([$modelName, 'factory'], []);
             foreach ($states as $state) {
-                $factory = $factory->$state();
+                if (method_exists(get_class($factory), $state)) {
```
shalvah, line 195: Why the wrapper here? What if the user made a typo in the state? Won't we be hiding errors?

shalvah:
Well, this seems like a lot of well-thought-out work (even though I don't fully understand it). Well done!

src/Tools/Utils.php
shalvah: Ah, I see. Because we're passing the states along in the recursive call, so they may/may not apply to the entire relation chain. Not a perfect solution, but it'll pass.

PR #375, opened 2021-11-30
Fix input name when the type of input is a list

AlirezaBarzin1:


shalvah:
Hmmm, does this work with Try It Out, via the lodash.set approach we're using?

shalvah:
Hmm...tested that this works with Try It Out, for both empty and filled states?

AlirezaBarzin1:
> Hmm...tested that this works with Try It Out, for both empty and filled states?

Yeah, I tested it out in both ways.

PR #376, opened 2021-12-02
Fix missing headings in sidebar

shalvah:


james2doyle:
Wow! Doesn't look at as bad you made it sound haha

I will take a look at the JS. I think it is just looking at the h2s

PR #377, opened 2021-12-02
Fixed: reduce margins and only pay attention to menu anchors

james2doyle:
Tried to fix the issues with the headers. Give this a shot now. The changes look at the nav to figure out how to track the titles. I also tweaked the ratio of the viewport. Might need some more tweaks but the logic is slightly better now

### DIFF
--- a/resources/js/theme-default.js
+++ b/resources/js/theme-default.js
@@ -14,7 +14,7 @@ document.addEventListener('DOMContentLoaded', function() {
 
         throttledUpdateHash(entries[0].target.id);
     }, {
-        rootMargin: '-8% 0px -8% 0px', // shrink the intersection viewport
+        rootMargin: '0px 0px -12% 0px', // shrink the intersection viewport
         threshold: 1.0, // trigger at 100% visibility
     });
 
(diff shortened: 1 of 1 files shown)


resources/js/theme-default.js
```diff
@@ -14,16 +14,21 @@ document.addEventListener('DOMContentLoaded', function() {
 
         throttledUpdateHash(entries[0].target.id);
     }, {
-        rootMargin: '-8% 0px -8% 0px', // shrink the intersection viewport
+        rootMargin: '0px 0px -12% 0px', // shrink the intersection viewport
```
james2doyle, line 17: I think we might need to use `rem` here in order to make all screens, no matter the size, work nicely. I will take a crack at that

shalvah:
Welp, deleted the branch accidentally. Please reopen to master.

shalvah:
Or never mind, pushed it back.

shalvah:
@james2doyle is there a reason we need the IntersectionObserver? Removed it and the browser handles the hash changes just fine. Removing it for now.

james2doyle:
@shalvah It was being used to track the scrolling so that the headers were triggering the menu items to become active

shalvah:
Oh...I see. Let's revisit that later, perhaps. Better to not have that extra feature, than to have a broken core feature (sidebar).

PR #379, opened 2021-12-05
Fixed: error when copying custom theme assets files

tee-moore:
i created a custom theme
but there was an error when running the command php artisan scribe:generate
error while copying css and js files to custom theme: "copy(public/docs/js/tryitout-3.16.0.js): failed to open stream: No such file or directory"

### DIFF
--- a/src/Writing/HtmlWriter.php
+++ b/src/Writing/HtmlWriter.php
@@ -76,10 +76,14 @@ class HtmlWriter
         Utils::copyDirectory("{$assetsFolder}/images/", "{$destinationFolder}/images");
 
         $assets = [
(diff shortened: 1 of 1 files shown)


shalvah:
Thank you, but this is not _the_fix. You're almost there, but you can't hardcode it to copy the `default` js and css. A better solution: in the `foreach` block some lines down, check if the file exists before trying to copy.

tee-moore:
sorry, maybe I don't quite understand ..
I hardcoded the default file, because the default file is always copied from the package, regardless of the name of the theme ..
if there is a variable then the file will not be copied for my theme ...

and in the foreach below, there is already a check if the file exists

shalvah:
Oh, right. I envisioned themes being provided by Scribe, but for user-defined themes, that will be a problem. Accepting this for now, but when we introduce more themes, we'll revisit.

Issue #380, opened 2021-12-06
Custom metadata strategy data not available in language templates


robbieaverill:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**
1. Create a custom metadata strategy
```php
<?php

namespace App;

use Knuckles\Camel\Extraction\ExtractedEndpointData;
use Knuckles\Camel\Extraction\Metadata;

class CustomMetadata extends Metadata
{
    public function __invoke(ExtractedEndpointData $endpointData, array $routeRules): array
    {
        return ['customMetadata' => true];
    }
}
```
2. Configure it:
```php
    /**
     * The strategies Scribe will use to extract information about your routes at each stage.
     * If you create or install a custom strategy, add it here.
     */
    'strategies' => [
        'metadata' => [
            Strategies\Metadata\GetFromDocBlocks::class,
            \App\CustomMetadata::class,
        ],
```
3. Override Javascript language template:
```php
# File: resources/views/vendor/scribe/partials/example-requests/javascript.blade.php
@php
    use Knuckles\Scribe\Tools\WritingUtils as u;
    /** @var  Knuckles\Camel\Output\OutputEndpointData $endpoint */

    dd($endpoint->metadata);
@endphp
```
4. Run `php artisan scribe:generate`

I expect to see `customMetadata: true` in the dumped endpoint metadata, however it's not there:

```
Knuckles\Camel\Extraction\Metadata^ {#967
  +groupName: null
  +beforeGroup: null
  +afterGroup: null
  +groupDescription: null
  +title: ""
  +description: ""
  +authenticated: false
  #exceptKeys: []
  #onlyKeys: []
}
```

**My environment:**
 - PHP version (from `php -v`): 8.0.5
 - Framework (Laravel/Lumen): Laravel
 - Laravel/Lumen version (from `composer show laravel/framework` or `composer show laravel/lumen-framework`): v8.74.0
 - Scribe version (from `composer show knuckleswtf/scribe`): 3.18.0

**My Scribe config (minus the comments):**

```php
<?php

use App\CustomMetadata;
use Knuckles\Scribe\Extracting\Strategies;

return [
    'strategies' => [
        'metadata' => [
            Strategies\Metadata\GetFromDocBlocks::class,
            CustomMetadata::class,
        ],
        'urlParameters' => [
            Strategies\UrlParameters\GetFromLaravelAPI::class,
[...]

PR #381, opened 2021-12-06
Add support for custom endpoint metadata definitions

robbieaverill:
Resolves https://github.com/knuckleswtf/scribe/issues/380

I'll make a separate PR for the docs

### DIFF
--- a/tests/Fixtures/TestCustomEndpointMetadata.php
+++ b/tests/Fixtures/TestCustomEndpointMetadata.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Knuckles\Scribe\Tests\Fixtures;
+
+use Knuckles\Camel\Extraction\ExtractedEndpointData;
+use Knuckles\Scribe\Extracting\Strategies\Strategy;
+
+class TestCustomEndpointMetadata extends Strategy
+{
+    public function __invoke(ExtractedEndpointData $endpointData, array $routeRules): ?array
+    {
+        $endpointData->metadata->custom['myProperty'] = 'some custom metadata';
+        return null;
+    }
+}
--- a/camel/Extraction/Metadata.php
+++ b/camel/Extraction/Metadata.php
@@ -2,7 +2,6 @@
 
 namespace Knuckles\Camel\Extraction;
 
-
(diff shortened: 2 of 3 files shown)


shalvah:
Very well done. But I'd rather avoid the magic ("any unknown properties go into `custom`"). For instance, we could add or rename a property in the future, and it would clash with something a user defined.

My idea of this was providing the `custom` property as a box for them to put whatever they wish in. I can see the difficulty with that, though, since returning something like this:

```php
return [
  "groupName" => "...",
  "custom" => [
    "a" => "b"
    ]
];
```

would overwrite the `custom` array. We could deep-merge it separately, but then we'd have to be sure we capture user's intent (unsetting a property, for instance). Maybe for now, we just keep it simple: add the `custom` property, and allow them to manipulate it directly?

```php
public function __invoke($endpointData)
{
  $endpointData->custom['a'] = 'b';
}
```


A second concern of mine is wrt serializing to the Camel files and loading from them. Did you test that the custom properties were serialized to the YAML as well, and properly loaded from them (can check this with `--no-extraction`)?

robbieaverill:
Sounds good, I'll update tomorrow

robbieaverill:
Hi @shalvah, I've updated to modify the custom property directly. I can confirm that YAML serialization and extraction is working as expected.

tests/Fixtures/TestCustomEndpointMetadata.php
```diff
@@ -0,0 +1,15 @@
...
+
+use Knuckles\Camel\Extraction\ExtractedEndpointData;
+use Knuckles\Scribe\Extracting\Strategies\Strategy;
+
+class TestCustomEndpointMetadata extends Strategy
+{
+    public function __invoke(ExtractedEndpointData $endpointData, array $routeRules): ?array
+    {
+        $endpointData->metadata->custom['myProperty'] = 'some custom metadata';
+        return null;
```
shalvah, line 13: I don't think this matters, does it? Functions in PHP implicitly return null.

Oh, this is because of the type annotation requiring a return statement.

shalvah:
3.19.0 '👍

PR #382, opened 2021-12-07
Add ability to define custom names for languages in config, by using the array key

robbieaverill:
This allows you to define custom language names which will be used in the output docs language switcher buttons, e.g.:

```php
# File: config/scribe.php
'example_languages' => [
    'bash',
    'javascript',
    'TS' => 'typescript',
],
```

Also fixes the language switcher on mobile which was broken (wouldn't switch language and would always send you to the top of the page).


shalvah:
3.19.0 '👍

PR #388, opened 2021-12-09
Use HTTPS for assets

HAstaShakiba:


shalvah:
I see you're a first-time contributor. Thank you for the PR👏

Unfortunately, this isn't exactly the problem/solution. Omitting the protocol tells the browser to use whatever protocol the page is on (see [this SO answer](https://stackoverflow.com/questions/15938970/what-can-go-wrong-with-scripts-that-have-no-protocol-in-their-src-url)). So, if your page is on `http`, the browser will fetch Lodash via `http`, and same for `https`.

shalvah:
On second thought...this means if you're viewing the docs by opening the file locally in your browser, the protocol will be `file://`, which means the browser will try to fetch `file://lodash`, which won't work. So maybe you're right, we should switch to `https`. No harm requesting HTTPS from a HTTP page.

shalvah:
Released in 3.19.1 👍

HAstaShakiba:
Thanks for the great package

PR #391, opened 2021-12-14
Feat: base url contain protocol

retry19:
# Overview

For now `baseUrl` not include protocol, only `localhost` or `example.com`, and in postman will be like this : 

```
http://{{baseUrl}}/some/other
```

It makes difficult when we want to change baseUrl from `http` to `https`, for example in local we use `http://localhost` but in staging we use `https://staging.example.com`. So, we must change environment variable in postman, then change `http` to `https` manually.

We can't only change environment variable, because `baseUrl` not contains protocol (http or https). It will be helpful when we can set `baseUrl` is http://localhost (including protocol).

So, I'm change baseUrl contain protocol, and also remove `protocol` attribute on `url` json.

# Example

## Before

```json
"variable": [
    {
        "id": "baseUrl",
        "key": "baseUrl",
        "type": "string",
        "name": "string",
        "value": "localhost"
    }
],
```

```json
"url": {
    "protocol": "http",
    "host": "{{baseUrl}}",
    "path": "api/withBodyParametersAsArray",
    "query": [],
    "raw": "http://{{baseUrl}}/api/withBodyParametersAsArray"
},
```
Screenshot (14)

## After

```json
"variable": [
    {
        "id": "baseUrl",
        "key": "baseUrl",
        "type": "string",
        "name": "string",
        "value": "http://localhost"
    }
],
```

```json
"url": {
    "host": "{{baseUrl}}",
    "path": "api/withBodyParametersAsArray",
    "query": [],
    "raw": "{{baseUrl}}/api/withBodyParametersAsArray"
},
```
Screenshot (13)

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -280,6 +280,15 @@ INTRO
         'overrides' => [
             // 'info.version' => '2.0.0',
         ],
+
+        /*
+         * Generate baseUrl variable contain protocol (http/https).
+         *
+         * Example :
+         * `true` -> baseUrl = http://localhost
+         * `false` -> baseUrl = localhost
+         */
+        'is_base_url_contain_protocol' => false,
     ],
 
     /*
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -280,15 +280,6 @@ INTRO
         'overrides' => [
             // 'info.version' => '2.0.0',
         ],
-
-        /*
-         * Generate baseUrl variable contain protocol (http/https).
-         *
-         * Example :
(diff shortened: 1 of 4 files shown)


shalvah:
Nice work. However, I won't be adding a new config item. Can you update the PR to make this the default instead? I don't think there's any advantage in removing the protocol from the base URL.

retry19:
> 

Okay, I'll update this PR

retry19:
@shalvah I already updated

PR #394, opened 2021-12-17
Updated OpenAPISpecWriter & Parameter classes to allow for the optional 'format' modifier field to be included in the output schema.

ArclightHub:
### What does this change do?

Updates the output openapi.yml file to allow the optional 'format' modifier as per the swagger data-types spec:
https://swagger.io/docs/specification/data-models/data-types/

### How does it work?
- Both of the Parameter classes have a new "public ?string $field = null;" property which can be optionally set to a string.
- The OpenAPISpecWriter class will include the above optional property in the 'schema' when it is set to a valid string.

### DIFF
--- a/camel/Extraction/Parameter.php
+++ b/camel/Extraction/Parameter.php
@@ -12,6 +12,7 @@ class Parameter extends BaseDTO
     public bool $required = false;
     public $example = null;
     public string $type = 'string';
+    public ?string $format = null;
 
     public function __construct(array $parameters = [])
     {
--- a/camel/Output/Parameter.php
+++ b/camel/Output/Parameter.php
@@ -19,6 +19,9 @@ class Parameter extends \Knuckles\Camel\Extraction\Parameter
     /** @var string */
     public string $type = 'string';
 
+    /** @var string|null */
+    public ?string $format = null;
+
     /** @var array */
     public $__fields = [];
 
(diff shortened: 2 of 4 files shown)


shalvah:
Sorry I didn't get to the opened issue earlier, but this doesn't sit well with me.

For one, adding a field to the `Parameter` DTO that is only relevant for the OpenAPIWriter does not seem right. The field is not used anywhere else in Scribe, and there isn't even a way to specify or hint its value via the default methods.

shalvah:
On the flip side, there's probably an argument to be made for adding a `custom` field to the `Parameter` class, [like we did for endpoint metadata](https://github.com/knuckleswtf/scribe/pull/381), where Scribe can put in extra data like an inferred `format`, but that's a thing for another day. 

I think what I'll go with for now is this:
1. I'll add a `custom` field to the `Parameter` class. You can put in whatever values you want here.
2. You can then create your own subclass of the OpenAPISpecWriter that reads the format from there, and bind to the container, and Scribe will [automatically pick that instead](https://github.com/knuckleswtf/scribe/blob/e43c36d24704b9e594cfbc25d67be94ffd69b009/src/Writing/Writer.php#L131).

shalvah:
v3.20 will contain the new `custom` field. You can write/read it however you wish. https://scribe.knuckles.wtf/laravel/reference/plugin-api#__invokeextractedendpointdata-endpointdata-array-routerules-array

PR #400, opened 2021-12-28
Fix bug where toggling the menu on mobile would jump to the top of the page

robbieaverill:


shalvah:
Not sure what this does exactly, but hope it works🤣

robbieaverill:
Hey @shalvah, what it's doing is preventing the default behaviour when you click on the "menu" button on mobile views. The button is an anchor, so the default behaviour is to follow the `href` attribute, which is `#`. What you saw before this patch is that clicking on the menu button on mobile would open the menu, and jump to the top of the page. It now only opens the menu.

shalvah:
Ah, thanks.

PR #408, opened 2022-01-20
Resolve issue with api resource

YannikFirre:
Hello,

I'm using laravel api resources and I solved my documentation generation issues with this PR.

I haven't tested in other cases at the moment.

Regards

### DIFF
--- a/camel/Extraction/ExtractedEndpointData.php
+++ b/camel/Extraction/ExtractedEndpointData.php
@@ -154,7 +154,7 @@ class ExtractedEndpointData extends BaseDTO
 
                 // We'll replace with {id} by default, but if the user is using a different key,
                 // like /users/{user:uuid}, use that instead
-                $binding = static::getFieldBindingForUrlParam($route, $singularResource, $singularResourceParam);
+                $binding = static::getFieldBindingForUrlParam($route, $singularResource, 'id');
 
                 if (!$foundResourceParam) {
                     // Only the last resource param should be {id}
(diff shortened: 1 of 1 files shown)


shalvah:
I don't really understand what's going on here. You have to provide a test case that describes what this is doing. If you don't know how to add a test to the project, you can at least post some example code here.

YannikFirre:
Yes sorry for the lack of information. 

A problem occurs when generating the documentation if the parameter contains underscore (like in api resource of Laravel). Instead of having an int, I get a string

Before: 
image

After:
image

shalvah:
I still need some more clarification, to be sure this doesn't break for others. How is your route declared? Something like this? `Route::resource("purpose-statuses")`? To confirm, you're saying that the route generated by Laravel uses underscores (_) in the URL parameter, even though you're using hyphen (-) in the path?

shalvah:
(Basically, a screenshot isn't enough. I need to see the code that led up to the screenshot.)

YannikFirre:
Yes this code 

```php
Route::apiResource('purpose-statuses', App\Http\Controllers\Property\PurposeStatusController::class);
```

Generate route like this : 
`api/purpose-statuses/{purpose_status}`

shalvah:
Right, thanks.

Next time, please add a test for this. I've added a test in https://github.com/knuckleswtf/scribe/commit/710dd8c7111a40c46d1a4364c83fac764bb79ba0 ([test results](https://github.com/knuckleswtf/scribe/runs/4926070836?check_suite_focus=true)), which should pass when this is merged. If you're confused about how to add tests, you can ask.

YannikFirre:
Thanks, I'll take a look =D

PR #409, opened 2022-01-20
ADDED - search id in DB when model instance

YannikFirre:
Hello,

This PR allows you to search for an existing ID in DB rather than using when no example.

Regards

### DIFF
--- a/src/Extracting/Strategies/UrlParameters/GetFromLaravelAPI.php
+++ b/src/Extracting/Strategies/UrlParameters/GetFromLaravelAPI.php
@@ -69,9 +69,14 @@ class GetFromLaravelAPI extends Strategy
                     $type = $this->normalizeTypeName($typeName);
                     $parameters[$paramName]['type'] = $type;
 
-                    $example  = $argumentInstance::first()->id ?? null;
+                    // Try to fetch an example ID from the database
+                    try {
+                        $example = $argumentInstance::first()->id ?? null;
+                    } catch (\Throwable $e) {
+                        $example = null;
+                    }
 
-                    if($example === null) {
+                    if ($example === null) {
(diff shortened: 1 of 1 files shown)


shalvah:
I don't really understand what's going on here. You have to provide a test case that describes what this is doing. If you don't know how to add a test to the project, you can at least post some example code here.

YannikFirre:
Yes sorry for the lack of information. 

When no example comment is specified, Scribe takes a randomly generated id which can cause a "ModelNotFoundException".

My modification allows to select the 1st id in database if it exists instead of taking a randomly generated id

src/Extracting/Strategies/UrlParameters/GetFromLaravelAPI.php
```diff
@@ -69,11 +69,15 @@ public function __invoke(ExtractedEndpointData $endpointData, array $routeRules)
                     $type = $this->normalizeTypeName($typeName);
                     $parameters[$paramName]['type'] = $type;
 
-                    // If the user explicitly set a `where()` constraint, use that to refine examples
-                    $parameterRegex = $endpointData->route->wheres[$paramName] ?? null;
-                    $example = $parameterRegex
-                        ? $this->castToType($this->getFaker()->regexify($parameterRegex), $type)
-                        : $this->generateDummyValue($type);
+                    $example  = optional($argumentInstance::first())->id;
```
shalvah, line 72: Use `??`, not `optional()`.

shalvah: One more thing: please wrap this in a try/catch, so it doesn't crash if the database isn't available.

YannikFirre:
Thanks, Sorry I forgot  =/

Issue #413, opened 2022-01-23
Adding additional meta for resource when calling with ApiResource


090809:
### Description
When I send response with Resource, sometimes I use `->additional()` call for adding meta information for display, like:

```
return EventResource::make($event)->additional([
    'title' => __('Event created'),
    'message' => __('Event successful created'),
]);
```

with current realization we can't add additional field to metadata

### Suggestion
Probably, we can extend `@apiResourceCollection` and `@apiResource` with key-value field, like:
`meta="title":"Event created"|"message":"Event successful created"` 

Or add this feature as additional tag like:
`@apiResourceAdditional title="Event Created" message="Event successful created" "my custom key"="my custom value"`

PR #414, opened 2022-01-24
[Feature] apiResourceAdditional tag, for documenting responses that using addtional function

090809:
ref: https://github.com/knuckleswtf/scribe/issues/413

### Description
When user send Resource with additional function, we can use `@apiResourceAdditional` tag for simultating constants for it

```
/**
 * @apiResource App\Http\Resources\EventResource
 * @apiResourceModel App\Models\Event
 * @apiResourceAdditional title="Event created" message="Event successful created"
 */
public function store(StoreEventRequest $request): JsonResponse
{
    // ... some create code ...
    return EventResource::make($event)->additional([
        'title' => 'Event created',
        'message' => 'Event successful created',
    ]);
}
```

### DIFF
--- a/src/Extracting/Strategies/Responses/UseApiResourceTags.php
+++ b/src/Extracting/Strategies/Responses/UseApiResourceTags.php
@@ -183,11 +183,9 @@ class UseApiResourceTags extends Strategy
      */
     private function getAdditionalData(?Tag $tag): array
     {
-        if (!$tag) {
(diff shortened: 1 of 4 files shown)


shalvah:
I generally don't like adding an extra annotation. Ideally this should be an attribute on the `@apiResourceModel`. but I can see the line getting too long, so I'll let this pass.

Issue #415, opened 2022-01-25
Bug with loading (creating) relation of belongsTo relation. (dot notation)


090809:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**
When using `@apiResourceModel` and `with` parameter, it WILL NOT create relationships of BelongsTo connected model.

Example:
I have an Event, it was created by user:
```php
class Event extends Model
{
    public function user(): BelongsTo
    {
        return $this->belongsTo(User::class);
    }
}
```

And I have a User model, that having MorphTo relation (spatie lib):
```php
class User extends Model
{
    public function media(): MorphMany
    {
        return $this->morphMany(config('media-library.media_model'), 'model');
    }
}
```

Of all of them I have an Factories. But!
For today, BelongsTo relations using this code:
https://github.com/knuckleswtf/scribe/blob/60e421cdded157864f8757878e23a144b7878169/src/Tools/Utils.php#L248-L256
So when Scribe calls ->create() of factory method, the Laravel will call this code [(link)](https://github.com/laravel/framework/blob/4735b0e287fa0fa35a7a98586497c89500c145a4/src/Illuminate/Database/Eloquent/Factories/Factory.php#L307-L318):

```php
protected function store(Collection $results)
{
    $results->each(function ($model) {
        if (! isset($this->connection)) {
            $model->setConnection($model->newQueryWithoutScopes()->getConnection()->getName());
        }

        $model->save();

        $this->createChildren($model);
    });
}
```

Where createChildren will call `createFor` as Parent instead of Child (there logic reverse), and it will exit from it without creating children of selected model, 'cause it doesn't have scenario for BelongsTo implementations
```php
    public function createFor(Model $parent)
    {
        $relationship = $parent->{$this->relationship}();
        
        if ($relationship instanceof MorphOneOrMany) { // its realizing by MorphOne and MorphMany
            $this->factory->state([
                $relationship->getMorphType() => $relationship->getMorphClass(),
                $relationship->getForeignKeyName() => $relationship->getParentKey(),
            ])->create([], $parent);
        } elseif ($relationship instanceof HasOneOrMany) { // its realizing by MorphOne, MorphMany, HasOne, HasMany
            $this->factory->state([
                $relationship->getForeignKeyName() => $relationship->getParentKey(),
            ])->create([], $parent);
[...]

Issue #412, opened 2022-01-22
Need to verbose exception, when ApiResourceModel factory creating failed


090809:
### Using:
* laravel: 8.80.0
* scribe: 3.21.0
* db: postgresql 12

### Description
We need to know, when factory fails with `->create()` method, cause it fails ALL transaction also.
I spent ~2 hours to find problem, why instansing with **fetch first** fails on selects with `In failed sql transaction: 7 ERROR:  current transaction is aborted, commands ignored until end of transaction block` message.

### Suggestion
Need to add verbose warn as here https://github.com/knuckleswtf/scribe/blob/5879aec913c4a4904446b6cfb748efbfc9d7ef90/src/Extracting/Strategies/Responses/UseTransformerTags.php#L168-L169
In:
https://github.com/knuckleswtf/scribe/blob/5879aec913c4a4904446b6cfb748efbfc9d7ef90/src/Extracting/Strategies/Responses/UseApiResourceTags.php#L190-L193
And in:
https://github.com/knuckleswtf/scribe/blob/5879aec913c4a4904446b6cfb748efbfc9d7ef90/src/Extracting/Strategies/Responses/UseTransformerTags.php#L163-L166

> p.s. Yeah, I know, that simple way was using additional DB, such as sqlite for these purposes, but I'm was too lazy to create all good factories for all models.

Issue #321, opened 2021-09-14
whenLoaded not working on api resource when using factories


denisgoryaynov:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

I have the following code in my api resource `UserResource`:
```
public function toArray($request): array
{
    $user = $this->user;

    return [
        'id' => $user->id,
        'name' => $user->name,
        'created_at' => $user->created_at,
        'updated_at' => $user->updated_at,
        'organization' => new OrganizationResource($this->whenLoaded('organization')), // this doesn't work with factory
        'organization' => new OrganizationResource($user->organization) // this works with factory, but I want to load my relations conditionally from my controller
   ];
}
```

And the following code in my `UserFactory`:
```
public function definition(): array
{
    return [
        'name' => $this->faker->name(),
        'created_at' => $this->faker->dateTime(),
        'updated_at' => $this->faker->dateTime(),
        'organization_id' => Organization::factory(),
    ];
}
```

And this is the comment in my `UserController`:
```
* @apiResource 200 App\Http\Resources\UserResource
* @apiResourceModel App\Models\User with=organization
```

What I expect from my resource is the following result:
```
{
    "data": {
        "id": 1,
        "name": "3d858092-e094-3d4e-8087-3b18dbec35b5",
        "created_at": "2009-08-04T17:42:10.000000Z",
        "updated_at": "1972-05-11T03:41:39.000000Z",
        "organization": {
            "id": 1,
            "name": "Murazik, Sporer and Glover",
            "created_at": "2021-09-14T14:52:07.000000Z",
            "updated_at": "2021-09-14T14:52:07.000000Z"
        }
    }
}
```

But what happens instead is I get:
```
{
    "data": {
        "id": 1,
        "name": "3d858092-e094-3d4e-8087-3b18dbec35b5",
        "created_at": "2009-08-04T17:42:10.000000Z",
        "updated_at": "1972-05-11T03:41:39.000000Z"
        // Organization is missing here
    }
}
```

If i use `$user->load('organization')` on the `$user` object created by the factory I can successfully see that the relationship is present inside the `relations` field of the Eloquent Model  and the `UserResource` successfully generates the expected output, but if I use `$user->with('organization')` on the `$user` object created by the factory the `relations` field is empty, so maybe it has something to do with the way factories load the relations in memory?

[...]

PR #417, opened 2022-01-26
Correct BelongsTo relation creation

090809:
Closes: #415
Closes: #412


090809:
@SBShane this test case needs to create :inmemory: db with migrations, so it failes. So here I'm stuck

090809:
Some nested data loading (in factory configuring) is workaround about #321 problem, so we can fix it clearly with real call of `->load($relations)` in `$factory->create()` (tested it with supposed test)

shalvah:
Hmm, I'll take a look at this and see what I can do.

shalvah:
Well done! I see you've put in a lot of work on the tests, but your change seems pretty innocuous, so I think we can live without the tests.

090809:
@shalvah You did not delete the `afterTestCleanup` function in tests, now it unused

shalvah:
Deleted it on master.

shalvah:
If we add more features that rely on the database, we may eventually add these database tests back, and your work might come in handy. Thanks!

PR #423, opened 2022-02-03
use default (Laravel URI) sorting for methods. Extend URI with http-method@path/to/route name

### LINKED ISSUE #389: Wrong alphabetical sort, when having many controllers in one group.
https://github.com/knuckleswtf/scribe/blob/30ea2dc97571b4ae675cb7f4ead6c0588d97a0ee/camel/Camel.php#L144

Example of current $e->endpointId()
```
GETapi-v1-industries
GETapi-v1-languages
```

in the above example all the get will be sorted together
Instead it should be $e->uri, so it can be sorted by the route itself. Example
```
api/v1/industries 
api/v1/industries/{industry} 
api/v1/languages
```

090809:
closes #389 

- [x] Rewrite ordering in tests

### DIFF
--- a/camel/Camel.php
+++ b/camel/Camel.php
@@ -126,18 +126,24 @@ class Camel
 
     /**
      * @param array[] $endpoints
+     * @param array $endpointGroupIndexes Mapping of endpoint IDs to their index within their group
      *
      * @return array[]
      */
-    public static function groupEndpoints(array $endpoints): array
+    public static function groupEndpoints(array $endpoints, array $endpointGroupIndexes): array
     {
         $groupedEndpoints = collect($endpoints)
             ->groupBy('metadata.groupName')
             ->sortKeys(SORT_NATURAL);
 
-        return $groupedEndpoints->map(function (Collection $endpointsInGroup) {
+        return $groupedEndpoints->map(function (Collection $endpointsInGroup) use ($endpointGroupIndexes) {
             /** @var Collection<(int|string),ExtractedEndpointData> $endpointsInGroup */
-            $sortedEndpoints = $endpointsInGroup->sortBy(fn(ExtractedEndpointData $e) => $e->uri);
+            $sortedEndpoints = $endpointsInGroup;
+            if (!empty($endpointGroupIndexes)) {
+                $sortedEndpoints = $endpointsInGroup->sortBy(
+                    fn(ExtractedEndpointData $e) => $endpointGroupIndexes[$e->endpointId()] ?? INF,
+                );
+            }
 
             return [
                 'name' => Arr::first($endpointsInGroup, function (ExtractedEndpointData $endpointData) {
--- a/camel/Camel.php
+++ b/camel/Camel.php
@@ -45,7 +45,7 @@ class Camel
      * Load endpoints from the Camel files into a flat list of endpoint arrays.
      *
      * @param string $folder
-     *
+     * @param bool $isFromCache
      * @return array[]
      */
     public static function loadEndpointsToFlatPrimitivesArray(string $folder, bool $isFromCache = false): array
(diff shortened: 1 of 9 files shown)


camel/Output/OutputEndpointData.php
```diff
@@ -137,7 +137,18 @@ public static function fromExtractedEndpointArray(array $endpoint): OutputEndpoi
 
     public function endpointId(): string
     {
-        return $this->httpMethods[0] . str_replace(['/', '?', '{', '}', ':', '\\', '+', '|'], '-', $this->uri);
+        return sprintf('%s@%s', $this->httpMethods[0], $this->uri);
+    }
+
+    public function cssEscapedEndpointId(): string
```
shalvah, line 143: I see what you're doing, but for consistency, why can't we just use the escaped version directly as the normal `endpointId()`? I'd prefer we not have two different types of endpoint IDs.

camel/Camel.php
```diff
@@ -126,24 +126,18 @@ public static function getEndpointIndexInGroup(array $groups, OutputEndpointData
...
         $groupedEndpoints = collect($endpoints)
             ->groupBy('metadata.groupName')
             ->sortKeys(SORT_NATURAL);
 
-        return $groupedEndpoints->map(function (Collection $endpointsInGroup) use ($endpointGroupIndexes) {
+        return $groupedEndpoints->map(function (Collection $endpointsInGroup) {
             /** @var Collection<(int|string),ExtractedEndpointData> $endpointsInGroup */
-            $sortedEndpoints = $endpointsInGroup;
-            if (!empty($endpointGroupIndexes)) {
-                $sortedEndpoints = $endpointsInGroup->sortBy(
```
shalvah, line 143: You shouldn't remove this condition. Sorting by `$endpointGroupIndexes` allows us to retain the user's sort order, if they have made changes to the Camel files. See https://scribe.knuckles.wtf/laravel/documenting/custom-endpoints#scenario-2-sorting-your-endpoints-or-groups

A better option would be to sort by `$endpointGroupIndexes` if present, or else sort by the URL.

shalvah (changes requested):

camel/Camel.php
shalvah: Also, I'm not even sure that we need this change. I don't think #389 is an issue at all. `$endpointGroupIndexes` is only present **if** you have run Scribe before. If you haven't, Scribe will not sort endpoints, which means you will get Laravel's default sort. Sorry, but I think this (and that issue) should be closed (unless someone can prove I'm wrong🤔).

The only issue I can see is that, if you add a new endpoint *after* running Scribe, it will automatically be put at the end of that group. But you can reset the sort order to the default with `--force`.

camel/Output/OutputEndpointData.php
090809: JS DOM finding will break with some characters, such as '@', ':', others, 'cause it might be css modifiers.

camel/Camel.php
090809: Yeah, I missed this part about Camel. But, unfortunately, sorting at the current moment  does not work "in alphabetical order".
Routes are generated from Router::getRoutes, which are not sorted. Therefore, normal (alphabetical) sorting does not happen.
I think we need to sort these paths in
https://github.com/knuckleswtf/scribe/blob/456d9204ef326366d0285d30d8db66197ab253fa/src/Matching/RouteMatcher.php#L21
Not after Camel processing.

shalvah: Alright, I see your point. But the PR is a bit to noisy for me to review. Please revert the endpointId changes, and add only the sorting.

Then, please add a small test that verifies that this sort order is kept, something like this:

```php
    /** @test */
    public function sorts_endpoints_by_uri()
    {
        RouteFacade::get('/api/industries', [TestController::class, 'dummy']);
        RouteFacade::get('/api/languages', [TestController::class, 'dummy']);
        RouteFacade::post('/api/languages', [TestController::class, 'dummy']);
        RouteFacade::get('/api/industries/{industry}', [TestController::class, 'dummy']);
        RouteFacade::get('/api/languages', [TestController::class, 'dummy']);
        RouteFacade::post('/api/industries/{industry}', [TestController::class, 'dummy']);
        RouteFacade::get('/api/languages/{language}', [TestController::class, 'dummy']);

        $this->artisan('scribe:generate');

        // Verify the order 
        $crawler = new Crawler(file_get_contents('public/docs/index.html'));
        $expectedEndpoints = $crawler->filter('h2');
        $this->assertEquals("GET./api/industries", $expectedEndpoints->getNode(0)->textContent);
        $this->assertEquals("GET./api/industries/{industry}", $expectedEndpoints->getNode(1)->textContent);
        $this->assertEquals("POST./api/industries/{industry}", $expectedEndpoints->getNode(2)->textContent);
        // ..and so on
    }
```

090809:
Sorry for lates, I will reimplement this as separeted PR-s later.

PR #424, opened 2022-02-03
fix(.editorconfig): added formatter for .json files

090809:


shalvah:
Hmm. I don't use .editorconfig (don't know if my IDE uses it), but I'll give this the benefit of the doubt.

PR #438, opened 2022-02-14
Fix validation rule accepted

hiszpek:
When using `accepted` or `accepted_if` validation rules on a field, the generated body parameters documentation hints that the value is an optional string. This is not true, as you can see in the Laravel Validation documentation: [https://laravel.com/docs/9.x/validation#rule-accepted](https://laravel.com/docs/9.x/validation#rule-accepted)

This PR aims to change the type, description and requirement of fields with those rules.

Documentation and example request generated before the changes:
image
image

Documentation and example request generated after the changes:
image
image

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -443,9 +443,7 @@ trait ParsesValidationRules
                     break;
                 case 'accepted_if':
                     $parameterData['type'] = 'boolean';
-                    $parameterData['description'] .= ' ' . $this->getDescription(
-                            $rule, [':other' => "<code>{$arguments[0]}</code>", ':value' => w::getListOfValuesAsFriendlyHtmlString(array_slice($arguments, 1))]
-                        ) . ' ';
+                    $parameterData['description'] .= " Must be accepted when <code>$arguments[0]</code> is " . w::getListOfValuesAsFriendlyHtmlString(array_slice($arguments, 1));
                     $parameterData['setter'] = fn() => true;
                     break;
                 case 'same':
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -188,6 +188,12 @@ trait ParsesValidationRules
                 case 'required':
                     $parameterData['required'] = true;
                     break;
+                case 'accepted':
+                    $parameterData['required'] = true;
+                    $parameterData['type'] = 'boolean';
+                    $parameterData['description'] .= ' Must be accepted.';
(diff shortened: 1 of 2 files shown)


src/Extracting/ParsesValidationRules.php
```diff
@@ -188,6 +188,12 @@ protected function parseRule($rule, array &$parameterData, bool $independentOnly
                 case 'required':
                     $parameterData['required'] = true;
                     break;
+                case 'accepted':
+                    $parameterData['required'] = true;
+                    $parameterData['type'] = 'boolean';
```
shalvah, line 193: This is good, but doesn't the documentation say it can be "yes" or 1? I don't think we can conclusively assign a `boolean` type.

hiszpek: Yes, you are right if we think in terms of strict typing. I wanted it to be consistent with the `boolean` [validation rule](https://laravel.com/docs/9.x/validation#rule-boolean), which can be true, false as well as 1, 0, "1" and "0" (but I see there's no "yes" and "no" option). It is still considered a `boolean` **type** in Scribe.

If we leave the type empty, it will default to `string` type, which I don't think makes much sense either, and from what I can see, displaying multiple available types would be pretty hacky.

hiszpek:
@shalvah I'm sorry, can you provide guidance on issues that popped up here?

I noticed that lint errors appeared before I started implementing thi PR. I can fix the issue about `Model::first()` being undefined (but I think it's outside of this PR's scope), but I don't really know how to fix the Filesystem errors, as I believe they are caused by code supporting versions 1 and 2, and I'm not really sure how to fix it other than adding `// @phpstan-ignore-line`.

shalvah:
Just ignore any lint errors that aren't from your code.

shalvah:
@hiszpek can you fix the tests?

hiszpek:
@shalvah The reason the tests on 7.4 fail is because it contains Laravel 6.20, but the `accepted_if` rule has been implemented somewhere around version 8. I couldn't find any information in contribution guide on how to handle situations like this, should I just scrap the `accepted_if` part and only leave the `accepted` rule? 

Sorry for leaving this hanging for a while.

shalvah:
https://github.com/knuckleswtf/scribe/search?q=marktestskipped

shalvah:
Feels like there's already an example for what to do in that file:

https://github.com/knuckleswtf/scribe/blob/04ac91453dae351d8f4652b852375429ba2b3b13/tests/Unit/ValidationRuleParsingTest.php#L323-L329

Issue #393, opened 2021-12-16
Uncaught TypeError: hljs.highlightAll is not a function


bjhijmans:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**
1. I updated Scribe from 3.16 to 3.19.1
2. Then I ran `php artisan scribe:generate`
3. But when I load the documentation page I got the error `Uncaught TypeError: hljs.highlightAll is not a function`

The error is on line 44 of theme-default-3.19.1.js. It didn't happen in 3.16, and I've been trying to understand why, but I can't figure it out. hljs is set, but it doesn't have the highlightAll function, even though highlight.js v10.7.2 does have that method.

**My environment:**
 - PHP 8.0.12
 - Framework Laravel:
 - Laravel version 8.76.1 
 - Scribe version 3.19.1: 

**My Scribe config (minus the comments):**
```
[
    'theme' => 'default',
    'title' => config('app.name') . ' API Documentation',
    'description' => '',
    'base_url' => null,
    'routes' => [[
        'match' => [
            'domains' => ['*'],
            'prefixes' => ['api/*'],
            'versions' => ['v1'],
        ],
        'include' => [],
        'exclude' => [],
        'apply' => [
            'headers' => ['Content-Type' => 'application/json', 'Accept' => 'application/json'],
            'response_calls' => [
                'methods' => [],
                'config' => ['app.env' => 'documentation'],
                'queryParams' => [],
                'bodyParams' => [],
                'fileParams' => [],
                'cookies' => [],
            ],
        ],
    ]],
    'type' => 'laravel',
    'static' => [
        'output_path' => 'public/docs',
    ],
    'laravel' => [
        'add_routes' => false,
        'docs_url' => '/api-documentation',
        'middleware' => [],
    ],

    'try_it_out' => [
        'enabled' => false,
        'base_url' => null,
        'use_csrf' => false,
        'csrf_url' => '/sanctum/csrf-cookie',
    ],
    'auth' => [
        'enabled' => true,
        'default' => true,
        'in' => 'bearer',
        'name' => 'key',
        'use_value' => env('SCRIBE_AUTH_KEY'),
        'placeholder' => '{API_TOKEN}',
        'extra_info' => '',
    ],
    'intro_text' => '...'
    'example_languages' => [
        'bash',
        'javascript',
        'python',
        'php',
    ],
    'postman' => [
[...]

PR #449, opened 2022-03-17
Update theme-default.js

abhaypokharkar:
New highlight js gives highlightAll not a function for laravel 9.


bjhijmans:
This is probably related to #393. My issue was caused by the Laravel Debugbar which includes an incompatible version of highlight.js, and shouldn't be enabled for the scribe docs.

shalvah:
Closing this PR because @abhaypokharkar has not sufficiently demonstrated the issue. This is lazy. You can't just say "the new highlight.js does this", so we should do this without looking at the version we're using. Scribe uses 10.x, but highlight/js' latest is 11.x.

PR #450, opened 2022-03-19
Feature/sort endpoints by title

negoziator:
### Description

This will change default sort behavior to sort all endpoints in each group by title ascending, by default.

### DIFF
--- a/camel/Camel.php
+++ b/camel/Camel.php
@@ -154,7 +154,7 @@ class Camel
                     })->metadata->groupDescription ?? '',
                 'endpoints' => $sortedEndpoints->map(
                     fn(ExtractedEndpointData $endpointData) => $endpointData->forSerialisation()->toArray()
-                )->values()->all(),
+                )->sortBy('metadata.title')->values()->all(),
             ];
         })->values()->all();
     }
--- a/src/Commands/GenerateDocumentation.php
+++ b/src/Commands/GenerateDocumentation.php
@@ -36,6 +36,8 @@ class GenerateDocumentation extends Command
     {
         // TODO Remove when Laravel 6 is no longer supported
         $this->getOutput()->write(str_repeat("\n", $count));
+
+        return $this;
     }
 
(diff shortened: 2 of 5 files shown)


shalvah:
...Why?

Sorting is a delicate thing. There have been a number of [issues and PRs](https://github.com/knuckleswtf/scribe/issues?q=sort) about it. There has to be a good justification for any change.

That said, I'm not convinced we should be sorting endpoints alphabetically. That can mess up a lot of things. Imagine having a group with: 
- Get a user (GET /users)
- Get a car (GET /cars)
- Buy a car (POST /cars/buy)
- Sell a car (POST /cars/sell)

WIth alphabetical sort, this would show up as:

- Buy a car (POST /cars/buy)
- Get a car (GET /cars)
- Get a user (GET /users)
- Sell a car (POST /cars/sell)

which is nonsense. I think sticking with the Laravel sort order (the order in which you added the routes) is much more sensible.

(Also: "alphabetical" order is a fickle thing when you think outside English.)

Maybe, to resolve sorting controversies, we should just provide a hook that lets you define your own sort order.🤔

negoziator:
Makes fine sense 😁

Thank you for taking your time on explaining in such details on howcome you find this a bad idea.

I completely get your point 👍🏼

negoziator:
@shalvah

What do you think about making it a configurable thing? 🤔

shalvah:
I opened an issue for it. I'll get to it later this week, but you can implement it if you wish. Basically, a new hook, like `beforeResponseCall()`. Would be pretty straightforward.

negoziator:
That's cool 😎

shalvah:
This is coming in v4. You'll be able to define an order in your config file, like this:


```
'groups' => [
        'order' => [
            // 'Group 1',
            // 'Group 2' => [
            //    'POST /this-endpoint-comes-first',
            //    'GET /this-endpoint-comes-next',
            // ]
        ],
]
```

PR #462, opened 2022-04-18
add assets prefix functionality

WordsofDefiance:
Hi! I needed this functionality for a project, and after some thought I concluded that the best way to implement it would be to actually add the functionality directly to Scribe.

This allows you to customize the location of the public assets generated for laravel blade documents. By default, the assets are created in `public/vendor/scribe`. With this feature, if the user defines a `laravel.asset_prefix` value, the path becomes `public/{{ asset_prefix }}/vendor/scribe`. 

I'm unclear how to best test this (my code currently passes all tests) so I'd be curious for ideas. Open to feedback.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -168,10 +168,10 @@ return [
         'docs_url' => '/docs',
 
         /*
-         * Prefix for public assets.
+         * Specify directory within `public` in which to store Laravel Scribe assets.
          * By default, assets are stored in `public/vendor/scribe`.
          */
-        'asset_prefix' => null,
+        'assets_directory' => null,
 
         /*
          * Middleware to attach to the docs endpoint (if `add_routes` is true).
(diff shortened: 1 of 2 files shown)


shalvah:
Hmmm, I'm reluctant to add this in this form. The Laravel convention is to place vendor assets in that `public/vendor`. If you want to change that, I think I'd favour an approach of customising the assets directory (within `public`) entirely, not adding a prefix.

WordsofDefiance:
I made that change you suggested. I renamed `asset_prefix` to `assets_directory`, and now the path itself is customizable (within the `public` directory).

shalvah:
Please send a PR to the docs repo as well, to update the config reference.

shalvah:
This page: https://github.com/knuckleswtf/scribe-docs/blob/master/laravel/reference/10-config.md

WordsofDefiance:
Here's the PR for documentation: 
https://github.com/knuckleswtf/scribe-docs/pull/14

PR #489, opened 2022-06-26
Fixed a bug in binding the URL parameters for API resource routes

### LINKED ISSUE #275: @urlParam as slug which is not primary key
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**

1. I have a model `Author` with a primary key as `int` `id` but I bind my routes with `string` `slug` to find details of model.

In `routes/api.php`, the URL parameter is called `author`.

```php
<?php

// ...
Route::get('/authors', [AuthorController::class, 'index'])->name('api.authors.index'); // authors list
Route::get('/authors/{author}', [AuthorController::class, 'show'])->name('api.authors.show'); // author show from author param as slug
```

In `app/Http/Controllers/Api:AuthorController.php`, I have to override with `@urlParam` for `id` instead of `author`, because scribe find `id` as primary key I suppose. It's not really a problem, except that seeing a parameter indicated as `id` can lead to errors but I don't know how to override the parameter name. I give an example for request to have correct response.

```php
<?php

/**
* Author details
*
* Details for one author, find by slug.
*
* @urlParam id string required The slug of author. Example: lovecraft-howard-phillips
*
*/
public function show(string $slug)
{
    try {
        $author = Author::whereSlug($slug)->with('media')->withCount('books')->firstOrFail();
        $author = AuthorResource::make($author);

        return $author;
    } catch (\Throwable $th) {
        return response()->json(['failed' => 'No result for '.$slug], 404);
    }
}
```

2. Then I ran `php artisan scribe:generate`, I've this result.

laravel-scribe-param-slug

3. But I saw, at right, I have correct `Example request` with `"http://localhost:8000/api/authors/lovecraft-howard-phillips"` but for the response, scribe try to find result with `id` `1` as parameter: `"failed": "No result for 1"`.

**My environment:**
 - PHP version: 8.0.7
 - Framework: Laravel v8.5.0
 - Scribe version: v3.6

<details>
  <summary>My Scribe config</summary>
  
```php
<?php

use Knuckles\Scribe\Extracting\Strategies;

return [
    'theme' => 'default',

    /*
     * The HTML <title> for the generated documentation. If this is empty, Scribe will infer it from config('app.name').
     */
    'title' => config('app.name').' API Documentation',

    /*
     * A short description of your API. Will be included in the docs webpage, Postman collection and OpenAPI spec.
     */
    'description' => 'The API documentation for '.config('app.name').' to use endpoints with another app.',

    /*
     * The base URL displayed in the docs. If this is empty, Scribe will use the value of config('app.url').
     */
    'base_url' => config('app.url'),

    /*
     * Tell Scribe what routes to generate documentation for.
     * Each group contains rules defining which routes should be included ('match', 'include' and 'exclude' sections)
     * and settings which should be applied to them ('apply' section).
     */
    'routes' => [
        [
            /*
             * Specify conditions to determine what routes will be a part of this group.
             * A route must fulfill ALL conditions to be included.
             */
            'match' => [
                /*
                 * Match only routes whose paths match this pattern (use * as a wildcard to match any characters). Example: 'users/*'.
                 */
                'prefixes' => ['api/*'],

                /*
                 * Match only routes whose domains match this pattern (use * as a wildcard to match any characters). Example: 'api.*'.
                 */
                'domains' => ['*'],

                /*
                 * [Dingo router only] Match only routes registered under this version. Wildcards are not supported.
                 */
                'versions' => ['v1'],
            ],

            /*
             * Include these routes even if they did not match the rules above.
             * The route can be referenced by name or path here. Wildcards are supported.
             */
            'include' => [
                // 'users.index', 'healthcheck*'
            ],

            /*
             * Exclude these routes even if they matched the rules above.
             * The route can be referenced by name or path here. Wildcards are supported.
             */
            'exclude' => [
                // '/health', 'admin.*'
            ],

            /*
             * Settings to be applied to all the matched routes in this group when generating documentation
             */
            'apply' => [
                /*
                 * Additional headers to be added to the example requests
                 */
                'headers' => [
                    'Content-Type' => 'application/json',
                    'Accept'       => 'application/json',
                ],

                /*
                 * If no @response or @transformer declarations are found for the route,
                 * Scribe will try to get a sample response by attempting an API call.
                 * Configure the settings for the API call here.
                 */
                'response_calls' => [
                    /*
                     * API calls will be made only for routes in this group matching these HTTP methods (GET, POST, etc).
                     * List the methods here or use '*' to mean all methods. Leave empty to disable API calls.
                     */
                    'methods' => ['GET'],

                    /*
                     * Laravel config variables which should be set for the API call.
                     * This is a good place to ensure that notifications, emails and other external services
                     * are not triggered during the documentation API calls.
                     * You can also create a `.env.docs` file and run the generate command with `--env docs`.
                     */
                    'config' => [
                        'app.env' => 'documentation',
                        // 'app.debug' => false,
                    ],

                    /*
                     * Query parameters which should be sent with the API call.
                     */
                    'queryParams' => [
                        // 'key' => 'value',
                    ],

                    /*
                     * Body parameters which should be sent with the API call.
                     */
                    'bodyParams' => [
                        // 'key' => 'value',
                    ],

                    /*
                     * Files which should be sent with the API call.
                     * Each value should be a valid path (absolute or relative to your project directory) to a file on this machine (but not in the project root).
                     */
                    'fileParams' => [
                        // 'key' => 'storage/app/image.png',
                    ],

                    /*
                     * Cookies which should be sent with the API call.
                     */
                    'cookies' => [
                        // 'name' => 'value'
                    ],
                ],
            ],
        ],
    ],

    /*
     * The type of documentation output to generate.
     * - "static" will generate a static HTMl page in the /public/docs folder,
     * - "laravel" will generate the documentation as a Blade view, so you can add routing and authentication.
     */
    'type' => 'static',

    /*
     * Settings for `static` type output.
     */
    'static' => [
        /*
         * HTML documentation, assets and Postman collection will be generated to this folder.
         * Source Markdown will still be in resources/docs.
         */
        'output_path' => 'public/docs',
    ],

    /*
     * Settings for `laravel` type output.
     */
    'laravel' => [
        /*
         * Whether to automatically create a docs endpoint for you to view your generated docs.
         * If this is false, you can still set up routing manually.
         */
        'add_routes' => true,

        /*
         * URL path to use for the docs endpoint (if `add_routes` is true).
         * By default, `/docs` opens the HTML page, `/docs.postman` opens the Postman collection, and `/docs.openapi` the OpenAPI spec.
         */
        'docs_url' => '/docs',

        /*
         * Middleware to attach to the docs endpoint (if `add_routes` is true).
         */
        'middleware' => [],
    ],

    'try_it_out' => [
        /*
         * Add a Try It Out button to your endpoints so consumers can test endpoints right from their browser.
         * Don't forget to enable CORS headers for your endpoints.
         */
        'enabled' => true,

        /*
         * The base URL for the API tester to use (for example, you can set this to your staging URL).
         * Leave as null to use the current app URL (config(app.url)).
         */
        'base_url' => null,
    ],

    /*
     * How is your API authenticated? This information will be used in the displayed docs, generated examples and response calls.
     */
    'auth' => [
        /*
         * Set this to true if any endpoints in your API use authentication.
         */
        'enabled' => false,

        /*
         * Set this to true if your API should be authenticated by default. If so, you must also set `enabled` (above) to true.
         * You can then use @unauthenticated or @authenticated on individual endpoints to change their status from the default.
         */
        'default' => false,

        /*
         * Where is the auth value meant to be sent in a request?
         * Options: query, body, basic, bearer, header (for custom header)
         */
        'in' => 'bearer',

        /*
         * The name of the auth parameter (eg token, key, apiKey) or header (eg Authorization, Api-Key).
         */
        'name' => 'key',

        /*
         * The value of the parameter to be used by Scribe to authenticate response calls.
         * This will NOT be included in the generated documentation.
         * If this value is empty, Scribe will use a random value.
         */
        'use_value' => env('SCRIBE_AUTH_KEY'),

        /*
         * Placeholder your users will see for the auth parameter in the example requests.
         * Set this to null if you want Scribe to use a random value as placeholder instead.
         */
        'placeholder' => '{YOUR_AUTH_KEY}',

        /*
         * Any extra authentication-related info for your users. For instance, you can describe how to find or generate their auth credentials.
         * Markdown and HTML are supported.
         */
        'extra_info' => 'You can retrieve your token by visiting your dashboard and clicking <b>Generate API token</b>.',
    ],

    /*
     * Text to place in the "Introduction" section, right after the `description`. Markdown and HTML are supported.
     */
    'intro_text' => <<<INTRO
This documentation aims to provide all the information you need to work with our API.

<aside>As you scroll, you'll see code examples for working with the API in different programming languages in the dark area to the right (or as part of the content on mobile).
You can switch the language used with the tabs at the top right (or from the nav menu at the top left on mobile).</aside>
INTRO
    ,

    /*
     * Example requests for each endpoint will be shown in each of these languages.
     * Supported options are: bash, javascript, php, python
     * To add a language of your own, see https://scribe.knuckles.wtf/laravel/advanced/example-requests
     *
     */
    'example_languages' => [
        'bash',
        // 'javascript',
    ],

    /*
     * Generate a Postman collection (v2.1.0) in addition to HTML docs.
     * For 'static' docs, the collection will be generated to public/docs/collection.json.
     * For 'laravel' docs, it will be generated to storage/app/scribe/collection.json.
     * Setting `laravel.add_routes` to true (above) will also add a route for the collection.
     */
    'postman' => [
        'enabled' => true,

        /*
         * Manually override some generated content in the spec. Dot notation is supported.
         */
        'overrides' => [
            // 'info.version' => '2.0.0',
        ],
    ],

    /*
     * Generate an OpenAPI spec (v3.0.1) in addition to docs webpage.
     * For 'static' docs, the collection will be generated to public/docs/openapi.yaml.
     * For 'laravel' docs, it will be generated to storage/app/scribe/openapi.yaml.
     * Setting `laravel.add_routes` to true (above) will also add a route for the spec.
     */
    'openapi' => [
        'enabled' => true,

        /*
         * Manually override some generated content in the spec. Dot notation is supported.
         */
        'overrides' => [
            // 'info.version' => '2.0.0',
        ],
    ],

    /*
     * Endpoints which don't have a @group will be placed in this default group.
     */
    'default_group' => 'Misc',

    /*
     * Custom logo path. This will be used as the value of the src attribute for the <img> tag,
     * so make sure it points to an accessible URL or path. Set to false to not use a logo.
     *
     * For example, if your logo is in public/img:
     * - 'logo' => '../img/logo.png' // for `static` type (output folder is public/docs)
     * - 'logo' => 'img/logo.png' // for `laravel` type
     *
     */
    'logo' => false,

    /*
     * If you would like the package to generate the same example values for parameters on each run,
     * set this to any number (eg. 1234)
     */
    'faker_seed' => null,

    /*
     * The strategies Scribe will use to extract information about your routes at each stage.
     * If you create or install a custom strategy, add it here.
     */
    'strategies' => [
        'metadata' => [
            Strategies\Metadata\GetFromDocBlocks::class,
        ],
        'urlParameters' => [
            Strategies\UrlParameters\GetFromLaravelAPI::class,
            Strategies\UrlParameters\GetFromLumenAPI::class,
            Strategies\UrlParameters\GetFromUrlParamTag::class,
        ],
        'queryParameters' => [
            Strategies\QueryParameters\GetFromFormRequest::class,
            Strategies\QueryParameters\GetFromInlineValidator::class,
            Strategies\QueryParameters\GetFromQueryParamTag::class,
        ],
        'headers' => [
            Strategies\Headers\GetFromRouteRules::class,
            Strategies\Headers\GetFromHeaderTag::class,
        ],
        'bodyParameters' => [
            Strategies\BodyParameters\GetFromFormRequest::class,
            Strategies\BodyParameters\GetFromInlineValidator::class,
            Strategies\BodyParameters\GetFromBodyParamTag::class,
        ],
        'responses' => [
            Strategies\Responses\UseTransformerTags::class,
            Strategies\Responses\UseResponseTag::class,
            Strategies\Responses\UseResponseFileTag::class,
            Strategies\Responses\UseApiResourceTags::class,
            Strategies\Responses\ResponseCalls::class,
        ],
        'responseFields' => [
            Strategies\ResponseFields\GetFromResponseFieldTag::class,
        ],
    ],

    'fractal' => [
        /* If you are using a custom serializer with league/fractal, you can specify it here.
         * Leave as null to use no serializer or return simple JSON.
         */
        'serializer' => null,
    ],

    /*
     * [Advanced] Custom implementation of RouteMatcherInterface to customise how routes are matched
     *
     */
    'routeMatcher' => \Knuckles\Scribe\Matching\RouteMatcher::class,

    /*
     * For response calls, API resource responses and transformer responses,
     * Scribe will try to start database transactions, so no changes are persisted to your database.
     * Tell Scribe which connections should be transacted here.
     * If you only use one db connection, you can leave this as is.
     */
    'database_connections_to_transact' => [config('database.default')],
];
```
  
</details>

Khaled-Farhat:
This pull request fixes #275. The bug is that Scribe uses the `id` as the default URL parameter when no inline binding is applied, but the user might be using other column by overriding the `getRouteKeyName()` method in the model.

### Before:
`\Camel\Extraction\ExtractedEndpointData@normalizeResourceParamName` will apply the inline binding if it exists, otherwise it will use the `id` as the default URL parameter.

### After:
1. `Camel\Extraction\ExtractedEndpointData@normalizeResourceParamName` will apply the inline binding if it exists, otherwise it will not alter the URL parameter name, as it is not its responsibility to guess the proper name.
2. Then, as the `UrlParameters\GetFromLaravelAPI` strategy will iterate over the method arguments, it can use the type-hinted variables to call the `getRouteKeyName()` method in the desired model. Then, it will bind the URL parameter name and set the parameter description.

**Note**: The current logic if there are neither inline binding nor type-hinted variables is to use the id, but the new logic in that case is to not alter the URL parameters. I updated the tests to use the new logic.

### DIFF
--- a/tests/GenerateDocumentationTest.php
+++ b/tests/GenerateDocumentationTest.php
@@ -415,6 +415,19 @@ class GenerateDocumentationTest extends BaseLaravelTest
     }
 
     /** @test */
+    public function generates_correct_url_params_using_bound_models()
+    {
+        RouteFacade::get('users/{user}', [TestController::class, 'withInjectedModel']);
+        config(['scribe.routes.0.match.prefixes' => ['*']]);
+        config(['scribe.routes.0.apply.response_calls.methods' => []]);
+
+        $this->artisan('scribe:generate');
+
+        $group = Yaml::parseFile('.scribe/endpoints/00.yaml');
+        $this->assertEquals('users/{id}', $group['endpoints'][0]['uri']);
+    }
+
+    /** @test */
     public function will_generate_without_extracting_if_noExtraction_flag_is_set()
     {
         config(['scribe.routes.0.exclude' => ['*']]);
--- a/tests/GenerateDocumentationTest.php
+++ b/tests/GenerateDocumentationTest.php
@@ -409,9 +409,9 @@ class GenerateDocumentationTest extends BaseLaravelTest
         $this->artisan('scribe:generate');
 
         $groupA = Yaml::parseFile('.scribe/endpoints/00.yaml');
-        $this->assertEquals('providers/{provider_slug}/users/{user_id}/addresses', $groupA['endpoints'][0]['uri']);
+        $this->assertEquals('providers/{provider_slug}/users/{user}/addresses', $groupA['endpoints'][0]['uri']);
         $groupB = Yaml::parseFile('.scribe/endpoints/01.yaml');
-        $this->assertEquals('providers/{provider_slug}/users/{user_id}/addresses/{uuid}', $groupB['endpoints'][0]['uri']);
+        $this->assertEquals('providers/{provider_slug}/users/{user}/addresses/{uuid}', $groupB['endpoints'][0]['uri']);
     }
 
     /** @test */
--- a/camel/Extraction/ExtractedEndpointData.php
+++ b/camel/Extraction/ExtractedEndpointData.php
@@ -137,7 +137,7 @@ class ExtractedEndpointData extends BaseDTO
         preg_match_all('#\{(\w+?)}#', $uri, $params);
 
         $resourceRouteNames = [
-            ".index", ".show", ".update", ".destroy",
+            ".index", ".store", ".show", ".update", ".destroy",
         ];
 
         if (Str::endsWith($route->action['as'] ?? '', $resourceRouteNames)) {
@@ -145,36 +145,21 @@ class ExtractedEndpointData extends BaseDTO
             $pluralResources = explode('.', $route->action['as']);
             array_pop($pluralResources);
 
-            $foundResourceParam = false;
+            $isLastResource = true;
             foreach (array_reverse($pluralResources) as $pluralResource) {
                 $singularResource = Str::singular($pluralResource);
                 $singularResourceParam = str_replace('-', '_', $singularResource);
 
-                $search = [
-                    "{$pluralResource}/{{$singularResourceParam}}",
-                    "{$pluralResource}/{{$singularResource}}",
-                    "{$pluralResource}/{{$singularResourceParam}?}",
-                    "{$pluralResource}/{{$singularResource}?}"
-                ];
+                $binding = static::getFieldBindingForUrlParam($route, $singularResourceParam);
 
-                // We'll replace with {id} by default, but if the user is using a different key,
-                // like /users/{user:uuid}, use that instead
-                $binding = static::getFieldBindingForUrlParam($route, $singularResource, 'id');
-
-                if (!$foundResourceParam) {
-                    // Only the last resource param should be {id}
-                    $replace = ["$pluralResource/{{$binding}}", "$pluralResource/{{$binding}?}"];
-                    $foundResourceParam = true;
-                } else {
-                    // Earlier ones should be {<param>_id}
-                    $replace = [
-                        "{$pluralResource}/{{$singularResource}_{$binding}}",
-                        "{$pluralResource}/{{$singularResourceParam}_{$binding}}",
-                        "{$pluralResource}/{{$singularResource}_{$binding}?}",
-                        "{$pluralResource}/{{$singularResourceParam}_{$binding}?}"
-                    ];
+                // If there is a field binding, like /users/{user:uuid}
+                if (!is_null($binding)) {
+                    // If the resource was the last, replace {singularResourceParam} with {binding}
+                    // otherwise, replace {singularResourceParam} with {singularResourceParam_binding}
+                    $uri = self::bindUrlParam($singularResourceParam, $binding, $isLastResource, $uri);
                 }
-                $uri = str_replace($search, $replace, $uri);
+
+                $isLastResource = false;
             }
         }
 
@@ -215,5 +200,31 @@ class ExtractedEndpointData extends BaseDTO
         }
 
         return $binding ?: $default;
+    }
+
+    public static function bindUrlParam(string $singularResourceParam, string $binding, bool $isLastResource, string $uri): string
+    {
+        $singularResource = str_replace('_', '-', $singularResourceParam);
+        $pluralResource = Str::plural($singularResource);
+
+        $search = [
+            "{$pluralResource}/{{$singularResourceParam}}",
+            "{$pluralResource}/{{$singularResource}}",
+            "{$pluralResource}/{{$singularResourceParam}?}",
+            "{$pluralResource}/{{$singularResource}?}"
+        ];
+
+        if ($isLastResource === true) {
+            $replace = ["$pluralResource/{{$binding}}", "$pluralResource/{{$binding}?}"];
+        } else {
+            $replace = [
+                "{$pluralResource}/{{$singularResource}_{$binding}}",
+                "{$pluralResource}/{{$singularResourceParam}_{$binding}}",
+                "{$pluralResource}/{{$singularResource}_{$binding}?}",
+                "{$pluralResource}/{{$singularResourceParam}_{$binding}?}"
+            ];
+        }
+
+        return str_replace($search, $replace, $uri);
     }
 }
--- a/src/Extracting/Strategies/UrlParameters/GetFromLaravelAPI.php
+++ b/src/Extracting/Strategies/UrlParameters/GetFromLaravelAPI.php
@@ -43,8 +43,16 @@ class GetFromLaravelAPI extends Strategy
         // and (User $user) model is typehinted on method
         // If User model has an int primary key, {user} param should be an int
 
+        // Also, bind url parameters for any bound models
+        // Eg Suppose route is /posts/{post},
+        // and (Post $post) model is typehinted on method
+        // If Post model has a slug routeKeyName, use {post_slug} parameter instead of {post}
+
         $methodArguments = $endpointData->method->getParameters();
-        foreach ($methodArguments as $argument) {
+        $currentArgumentPosition = 0; // to check if we are processing the last argument
+        foreach (array_reverse($methodArguments) as $argument) {
+            ++$currentArgumentPosition;
+
             $argumentType = $argument->getType();
             // If there's no typehint, continue
             if (!$argumentType) {
@@ -54,10 +62,31 @@ class GetFromLaravelAPI extends Strategy
                 $argumentClassName = $argumentType->getName();
                 $argumentInstance = new $argumentClassName;
                 if ($argumentInstance instanceof Model) {
+                    $routeKeyName = $argumentInstance->getRouteKeyName();
+
                     if (isset($parameters[$argument->getName()])) {
-                        $paramName = $argument->getName();
-                    } else if (isset($parameters['id'])) {
-                        $paramName = 'id';
+                        $oldParamName = $argument->getName();
+
+                        $isLastResource = ($currentArgumentPosition === 1);
+                        $endpointData->uri = $endpointData->bindUrlParam($oldParamName, $routeKeyName, $isLastResource, $endpointData->uri);
+
+                        // Find the new paramName
+                        if ($isLastResource === true) {
+                            // We are processing the last argument, use {routeKeyName} as field binding
+                            $paramName = $routeKeyName;
+                        } else {
+                            // We are processing an earlier argument, use {argumentName_routeKeyName} as field binding
+                            $paramName = $argument->getName() . '_' . $routeKeyName;
+                        }
+
+                        // Rename the parameter from oldParamName to paramName
+                        $parameters[$paramName] = $parameters[$oldParamName];
+                        $parameters[$paramName]['name'] = $paramName;
+                        unset($parameters[$oldParamName]);
+
+                        $parameters[$paramName]['description'] = $this->inferUrlParamDescription($endpointData->uri, $paramName, $oldParamName);
+                    } else if (isset($parameters[$routeKeyName])) {
+                        $paramName = $routeKeyName;
                     } else {
                         continue;
                     }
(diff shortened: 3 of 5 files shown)


tests/GenerateDocumentationTest.php
```diff
@@ -409,9 +409,22 @@ public function generates_correct_url_params_from_resource_routes_and_field_bind
         $this->artisan('scribe:generate');
 
         $groupA = Yaml::parseFile('.scribe/endpoints/00.yaml');
-        $this->assertEquals('providers/{provider_slug}/users/{user_id}/addresses', $groupA['endpoints'][0]['uri']);
+        $this->assertEquals('providers/{provider_slug}/users/{user}/addresses', $groupA['endpoints'][0]['uri']);
```
shalvah, line 412: The problem is that we don't want this. This sued to be the case before, but a number of people pointed out that it was unintuitive. It makes sense from the Laravel perspective, since we know about route model binding, but the reader of the API docs wants to know the value that goes in there (the user ID), not just the model.

Khaled-Farhat: I agree that using only the model name is not intuitive, but that is not the case when using type-hinted variables in the controller action.
In the test case you pointed out, there are no type-hinted variables in the controller action, and that is the reason why the URL parameter was `user`.
But when using type-hinted variables in the controller action, the URL parameter will be `user_id` (or `user_slug` if the developer customized the route key name to `slug`). I think this behavior satisfies both the API docs reader and the developer since he can customize the route key name.

shalvah: Ah, I don't agree there. 

> since he can customize the route key name

Yes, but will he? It would be annoying to ask a developer to go through their routes and add `:id` for every parameter they use. Especially as Laravel defaults to using the `id` as the route key, it makes no sense that we can't.

shalvah: From your original comment:

> The bug is that Scribe uses the id as the default URL parameter when no inline binding is applied, but the user might be using other column by overriding the getRouteKeyName() method in the model.

I think we should focus on this. See if the model has `getRouteKeyName()` defined or a custom key in the route, and use that. Otherwise, we stick to `id`. We may just need to move the order of operations around to achieve this.

Khaled-Farhat: So you want to stick with the `id` even for varaibles that are not type-hinted? as the `id` will be used by default for the type-hinted variables because `getRouteKeyName()` is defined in `Illuminate\Database\Eloquent\Model` to return the `id`.

shalvah: When you say "variables that are not type-hinted", what fo you mean?

Khaled-Farhat: I mean the method arguments of the action that have types (`User $user` for example).

shalvah: If you're talking about a regular URL parameter, it doesn't get bound to a model by Laravel, and we don't touch it either. For instance, `/things/{thing}` should be as-is, unless it's a resource route or there's a `Thing $thing` parameter defined.

Khaled-Farhat: Ok, I see your point.

As I understand, the required behavior is:
For non-resource routes, keep the current behavior of Scribe.
For resource routes:
- If there is a custom key in the route, use it,
- Else, if the method argument in the action is type-hinted call `getRouteKeyName()` in the model.
- Otherwise, use the `id`.

I will give it a try. Do you prefer I make the changes in a new clean pull request?

Khaled-Farhat: Could you please confirm that this is the intended behavior?

shalvah: Could you give examples? It's been a while since I wrote this code, and I don't remember exactly the scenarios. It's easier to talk in code examples. Just short examples of a route for each case you've mentioned will do.

Khaled-Farhat: Sure. The current behavior:
1. If there is an inline binding in the route, use it.
2. Else, if it is a resource route parameter, use the id.
3. Otherwise, keep the same parameter name.

What I will do in my new pull request, is to add a new step before the second step to check if there is a type-hinted variable in the controller action with the same name as the route segment name. This is the same behavior as the laravel [implicit binding](https://laravel.com/docs/9.x/routing#implicit-binding):

> Laravel automatically resolves Eloquent models defined in routes or controller actions whose type-hinted variable names match a route segment name.
---

#### Case 1: If there is an inline binding in the route.
```
Route::apiResource('posts', PostController::class)->only('show')->parameters(['posts' => 'post:slug']);
Route::get('posts/{post:slug}/comments', function() {});
```
Current behavior: use inline binding. Current URIs: `posts/{slug}` and `posts/{post_slug}/comments`.
The new behavior is the same.

#### Case 2: No inline binding exists, but model binding exists.
Assume that `Post@getRouteKeyName` returns `slug`, and `PostController@show` uses model binding (`Post $post`).
```
Route::apiResource('posts', PostController::class)->only('show').
Route::get('posts/{post}/comments', function(Post $post) {});
```
Current behavior: use the `id` for resource parameters, and keep the same name for non-resource parameters.
Current URIs: `posts/{id}` and `posts/{post}/comments`.
New behavior: call `getRouteKeyName()`.
New URIs: `posts/{slug}` and `posts/{post_slug}/comments`.

#### Case 3: Resource route parameter. No inline binding exists. No model binding exists.
Assume that `PostController@show` does not use model binding.
```
Route::apiResource('posts', PostController::class)->only('show');
```
Current behavior: use the `id`. Current URI: `posts/{id}`.
The new behavior is the same.

#### Case 4: Non-Resource route. No inline binding exists. No model binding exists.
```
Route::get('posts/{post}/commets', function($postId) {});
````
Current behavior: keep the same parameter name. Current URI: `posts/{post}`.
The new behavior is the same.

Khaled-Farhat: I have another question about the testing, Is there a problem if I added 3 fixtures to test the new behavior?

The fixtures I want to add:
- `TestPost` model that overrides `getRouteKeyName()` to return `slug`.
- `TestPostController` with one method that has a type-hinted variable (`Post $post`).
- `TestPostUserController` with one method that uses two type-hinted variables (`Post $post, User $user`). I will use it to test nested resource routes.

shalvah: Yes, that sounds good. We probably need to clean up the code to make it less confusing, perhaps with some inline documentation of the examples. Also, I think the responsibility is half-shared; IIRC, there's somewhere else in the code where the URI is "fixed" slightly, which may change the results from here.

shalvah: Yeah, you can add whatever you need.

PR #494, opened 2022-07-09
Fix documentation generation when using binded interfaces instead of model classes as typed arguments

toyi:
Hi,

The version `3.33.1` introduces a breaking change when you are using a type hinted argument that is an interface binded to a model via the service container.

Since an interface is not instantiable, you get this error when generating the documention:

```
 Error 

  Cannot instantiate interface App\Interfaces\YourModelInterface.php

  at vendor/knuckleswtf/scribe/camel/Extraction/ExtractedEndpointData.php:262
    258▕             // The argument does not have a type-hint
    259▕             return false;
    260▕         } else {
    261▕             $argumentClassName = $argumentType->getName();
  ➜ 262▕             $argumentInstance = new $argumentClassName;
    263▕             return ($argumentInstance instanceof Model);
    264▕         }
    265▕     }
    266▕ }

      +21 vendor frames 
  22  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```

To resolve this issue, we need to use the service container to resolve the instance instead of just `new`.

However, there is a caveat, because when you resolve a form request class using the service container, it also execute the `authorize` method of the `ValidatesWhenResolvedTrait` trait.

That was the first error I found but alongs the way of fixing it, I found that type hinted arguments with builtin types (e.g. `string`) also returns a similar error:

```
 Error 

  Class "string" not found

  at vendor/knuckleswtf/scribe/camel/Extraction/ExtractedEndpointData.php:262
    258▕             // The argument does not have a type-hint
    259▕             return false;
    260▕         } else {
    261▕             $argumentClassName = $argumentType->getName();
  ➜ 262▕             $argumentInstance = new $argumentClassName;
    263▕             return ($argumentInstance instanceof Model);
    264▕         }
    265▕     }
    266▕ }

      +21 vendor frames 
  22  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```

--------------

This PR fixes these problems by using the `app()` helper to resolve any interfaces that can be used as controller method arguments, `new` for classes (to avoid any side effects of `ValidatesWhenResolvedTrait`).

Althrough this PR also fixes the error for builtin types, it was not its first goal. For now, theses arguments are treated as "not type hinted".

Of course, any improvements are welcome :)

Have a nice day.

### DIFF
--- a/camel/Extraction/ExtractedEndpointData.php
+++ b/camel/Extraction/ExtractedEndpointData.php
@@ -267,7 +267,7 @@ class ExtractedEndpointData extends BaseDTO
     protected function argumentHasModelType(\ReflectionParameter $argument): bool
     {
         $argumentType = $argument->getType();
-        if (!$argumentType) {
+        if (!$argumentType instanceof \ReflectionNamedType) {
             // The argument does not have a type-hint
             return false;
         } else {
--- a/camel/Extraction/ExtractedEndpointData.php
+++ b/camel/Extraction/ExtractedEndpointData.php
@@ -267,7 +267,7 @@ class ExtractedEndpointData extends BaseDTO
     protected function argumentHasModelType(\ReflectionParameter $argument): bool
     {
         $argumentType = $argument->getType();
-        if (!$argumentType instanceof \ReflectionNamedType) {
+        if (!($argumentType instanceof \ReflectionNamedType)) {
             // The argument does not have a type-hint
             return false;
         } else {
--- a/camel/Extraction/ExtractedEndpointData.php
+++ b/camel/Extraction/ExtractedEndpointData.php
@@ -210,7 +210,21 @@ class ExtractedEndpointData extends BaseDTO
         return $copy;
     }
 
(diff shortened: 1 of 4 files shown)


camel/Extraction/ExtractedEndpointData.php
```diff
@@ -254,12 +267,11 @@ protected function getTypeHintedArguments(): array
     protected function argumentHasModelType(\ReflectionParameter $argument): bool
     {
         $argumentType = $argument->getType();
-        if (!$argumentType) {
+        if (!$argumentType instanceof \ReflectionNamedType) {
```
shalvah, line 270: Shouldn't this be


```suggestion
        if (!($argumentType instanceof \ReflectionNamedType)) {
```

shalvah (approved):

shalvah:
Thanks for the catch and the fix!

toyi:
@shalvah My pleasure :) However I just realized I did an awful english mistake, everywhere I wrote "binded", it should be "bound". Can't make a fix for this typo right now but I'll do it tomorrow if you didn't already.

PR #505, opened 2022-08-06
Add resources response field tags to documentation

N0rb3Rt88:
Using laravel framework looks easier to add @responseField annotation to resource once than repeat the same annotations over and over in controller. So if in controller exists @apiResource or @apiResourceCollection annotation, Scribe will merge the resource toArray method's annotations with controller method annotations.

Another small change, that if base_path method exists, Scribe will prepare laravel type output path with this method. It was necesarry for me when the generation process started from controller or from a job and not from console.

I hope these changes can be useful for others too.


lizardopc:
this broke the use of attributes in @apiResource

`/**
 * @apiResource 201 App\Http\Resources\UserResource
 * @apiResourceModel App\Models\User
 */`

this 

` $className = $apiResourceTag->getContent();`

returns

201 App\Http\Resources\AccountResource

and then

`  Class "201 App\Http\Resources\AccountResource" does not exist

  at vendor/knuckleswtf/scribe/src/Tools/Utils.php:208
`

PR #529, opened 2022-10-07
Adding support of global bearer token variable.

buglinjo:
This update brings global bearer token setter in the postman file. If you set `use_value` as `{{token}}` in the scribe config file it'll automatically populate it when postman file is generated.

In you postman variables you can set that token automatically when sending login request using Postman's `Tests`. You can set this JS code in Tests tab and JWT token will be automatically populated and used globally by the API.
```javascript
var jsonData = JSON.parse(responseBody);
pm.environment.set("token", jsonData.login.access_token);
```

### DIFF
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -77,6 +77,13 @@ class PostmanCollectionWriter
             ],
             "bearer" => [
                 'type' => 'bearer',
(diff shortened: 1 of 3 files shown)


shalvah:
Update the test fixtures^^^

buglinjo:
Updated Fixtures/collection.json and postman file generation test to support bearer global value

PR #556, opened 2022-11-14
Check and generate ->data on generateResponseContentSpec()

Maras0830:
Hi scribe,

I found some issue about generate OpenAPI format.

when I use `ResponseFromFile` and write `ResponseField` on my controller 

ResponseFromFile file:
```json
{
    "data": [
        {
            "name": "Maras",
            "id": 1,
        }
    ],
   "status": 200
}
```

ResponseField:
```
#[ResponseField('name', 'string', 'User Name')]
#[ResponseField('id', 'integer', 'User unique number')]
```

and i write columns via ResponseField that will return:
```yaml
                properties:
                  data:
                    type: array
                    example:
                      -
                        name: Maras
                        id: 1
```

but i read [official document](https://scribe.knuckles.wtf/laravel/documenting/responses#response-fields), the document says
```
You don't need to specify the full field path if the field is inside an array of objects or wrapped in pagination data. For instance, the above annotation will work fine for all of these responses:
```

so i think properties should return:
```yaml
 properties:
                  name:
                    type: string
                    example: Maras
                    description: 'User Name'
                  id:
                    type: integer
                    example: 1
                    description: 'User unique number'
```

### DIFF
--- a/src/Writing/OpenAPISpecWriter.php
+++ b/src/Writing/OpenAPISpecWriter.php
@@ -378,6 +378,14 @@ class OpenAPISpecWriter
                 ];
 
             case 'object':
+                if (property_exists($decoded, 'data')) {
+                    $decoded = $decoded->data;
+                }
+
+                if (is_array($decoded)) {
+                    $decoded = Arr::first($decoded);
+                }
+
                 $properties = collect($decoded)->mapWithKeys(function ($value, $key) use ($endpoint) {
                     $spec = [
                         // Note that we aren't recursing for nested objects. We stop at one level.


shalvah:
Hm, I'm confused. There doesn't seem to be anything wrong here. Your response from your file has the fields nested in a `data` key, so this JSON schema is correct:


```tml
    properties:
                  data:
                    type: array
                    example:
                      -
                        name: Maras
                        id: 1
```

What you're proposing describes a response like this, 

```json
{
  name": "Maras",
  id": 1,
}
```
which isn't what you have.

Maras0830:
> ```yaml
> properties:
>                   name:
>                     type: string
>                     example: Maras
>                     description: 'User Name'
>                   id:
>                     type: integer
>                     example: 1
>                     description: 'User unique number'
> ```

The point of the question is the `properties`, every return columns should have `type` & `description`, but that just return example doesn't have any `type` & `description` on origin generator result.

``yaml
properties:
                  name:
                    type: string
                    example: Maras
                    description: 'User Name'
                  id:
                    type: integer
                    example: 1
                    description: 'User unique number'
```

shalvah:
True, but this is still not correct. You can't put the fields directly on `properties`. Then you're describing a totally different shape from what you actually have there.

shalvah:
Closing this PR because your solution is fundamentally wrong. We can still discuss the issue, but your approach is wrong—it does not try to determine the user's intent and simply assumes a `data` field must mean what you think it is. What if it's a regular object that just happens to have a `data` field?

Also, it does not have tests, especially for such an important change.

shalvah:
Your approach is both wrong in intention (wrongly interprets some cases) and produces a wrong output spec, even when it's the right intention. Thanks for trying, but there's a better way to achieve this. I'll create an issue for it (but will not get around to it for a long time).

PR #577, opened 2022-12-05
Feature:document generation support i18n

fplun:
Add configuration item:language

Support multi language conversion to ASCII when generating documents

### DIFF
--- a/camel/Output/OutputEndpointData.php
+++ b/camel/Output/OutputEndpointData.php
@@ -154,7 +154,7 @@ class OutputEndpointData extends BaseDTO
 
     public function fullSlug(): string
     {
-        $groupSlug = Str::slug($this->metadata->groupName);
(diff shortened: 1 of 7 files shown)


shalvah:
Thanks for the PR. I'm a bit dissatisfied, though.
1. The config file param `language` does not explain what it's for. This is misleading because of point (2):
2. This isn't full internationalization. Which is okay, but I'd rather release internationalization as a complete feature. Having `language` in there when it's only used for generating slugs would definitely be confusing. Perhaps I'll create a branch for that.
3. Also, it looks like your usage of `Str::slug` is incorrect? `language` is the third parameter: https://github.com/illuminate/support/blob/475bda9d24e832388d5514683bec5ea6ef9a3c94/Str.php#L1016

I think it's a good start, though.

fplun:
My English is poor，I'm looking forward to it. I hope this project will be better and better

PR #586, opened 2022-12-12
Remove redundant imports

AdamiecRadek:
This PR removes two redundant imports from `src/Commands/GenerateDocumentation.php` that were not used and one import of a non-existing class in `src/Writing/OpenAPISpecWriter.php`


AdamiecRadek:
It's strange because I did not touch the file that was failing the lint check, but I fixed it anyway :)

PR #591, opened 2022-12-21
Introduce PHP CS fixer tool and apply PSR standards

AdamiecRadek:
This PR introduces [PHP CS Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer) tool into Scribe.   

**I divided PR into smaller commits to make code review easier**

The most important part of this PR is the `.php-cs-fixer.php` file which contains all the rules for the cs fixer. I added there PSR12 and PHP80 rulesets.    

I also updated the composer scripts' names to reflect better what they are doing. I added a new composer script to git flow.    

Everything else is a result of an automatic code-style fix.


shalvah:
Whoops, I'm really sorry I have to close this after the work you've put in, but I don't like style-checking tools. I won't go into details why, but they're even more of a nuisance on a project like this—I'm by far the main contributor, and any weird styles I don't like can be fixed pretty easily. Style fixers should be introduced when there is a problem (multiple primary contributors with differing code styles and significant time wasted on style debates), and there's no such problem here.

So, yeah—sorry, but closing this. Thanks for the contribution attempt!

AdamiecRadek:
I appreciate your honest and detailed answer. I may have a different view on the matter, but I understand your reasoning, so no worries :)

PR #595, opened 2022-12-29
Allow changing the version number within the OpenApi-Specs

Yasu-San:
Adds the possibility to change the Version number within the generated OpenAPI-Document


shalvah:
Sorry, this does not meet the threshold for a new config item. Use `openapi.overrides` if you wish to change the version.

Issue #455, opened 2022-03-29
array to string conversion 


ibrahem-kamal:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

when adding nested array it fails in generating the docs and gives the error below , 

<img width="954" alt="image" src="https://user-images.githubusercontent.com/38753243/160589412-dc31449f-a930-44a0-854e-3c1dd22372e4.png">

it fails because of the filter.categories_ids and filter.brands_ids are arrays 

<img width="649" alt="image" src="https://user-images.githubusercontent.com/38753243/160589392-ef21d4aa-c206-4c2e-9406-34613b4b13cd.png">

issakujitsuk:
I'm facing a similar problem, so I'll add some information.

Points:
* 'example_languages': "bash", "javascript", "php", etc
* Route method must be "get"

## Case 1.

routes/api.php:
```php
/**
 * ?param[id][0]=1
 * @queryParam param object
 * @queryParam param.id integer[] required
 */
Route::get("/test", function () {
    return "Test";
});
```
Then
```
  at vendor/knuckleswtf/scribe/src/Tools/WritingUtils.php:119
    115▕                 } else {
    116▕                     // Hash query param (eg filter[name]=john should become "filter[name]": "john")
    117▕                     foreach ($value as $item => $itemValue) {
    118▕                         $output .= str_repeat(" ", $spacesIndentation);
  ➜ 119▕                         $output .= "$startLinesWith$quote$parameter" . "[$item]$quote$delimiter $quote$itemValue$quote$endLinesWith\n";
    120▕                     }
    121▕                 }
    122▕             }
    123▕         }

      +9 vendor frames 
  10  storage/framework/views/a445359943839067f85d5861bd9b00bd1213813a.php:15
      Knuckles\Scribe\Tools\WritingUtils::printQueryParamsAsKeyValue([], "'", "=>", "[]")

      +8 vendor frames 
  19  storage/framework/views/82d66105162ba1e5469d368df4c08c33adede6c9.php:21
      Illuminate\View\View::render()
```

## Case 2.

routes/api.php:
```php
/**
 * ?param[0][id]=1
 * @queryParam param object[]
 * @queryParam param.*.id integer required
 */
Route::get("/test", function () {
    return "Test";
});
```
Then
```
  at vendor/knuckleswtf/scribe/src/Tools/WritingUtils.php:114
[...]

Issue #308, opened 2021-08-30
Problem with generating sample when using file type in nested arrays in validation rules


MohammadMirsafaei:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**
1. I set my configuration to default Scribe config.
2. Then I ran `php artisan scribe:generate`
3. But I saw that for my endpoint it couldn't generate example `body` request in javascript.
it show only: 
`let body = `

instead of
```
cosnt body = new FormData();
body.append....
```

**Screenshots and stack traces:**

**My environment:**
 - PHP version (from `php -v`):  7.3.29
 - Framework (Laravel/Lumen): Laravel
 - Laravel/Lumen version (from `composer show laravel/framework` or `composer show laravel/lumen-framework`): 8.52.0
 - Scribe version (from `composer show knuckleswtf/scribe`): 2.7.10

**My Scribe config (minus the comments):**
It's Scribe default config
**Additional info:**
My validation rules:
```
'data' => 'filled|array',
            'data.*' => 'required|array',
            'data.*.is_file' => 'required|boolean',
            'data.*.value' => 'filled|string|required_without:data.*.file',
            'data.*.file' => 'filled|required_without:data.*.value|file|mimes:png,jpg,gif,mp4,mp3|mimetypes:image/jpeg,image/png,video/mp4,audio/mpeg',
```
I think the problem is when using 2 nested array and the type is `file`.
If I remove `data.*.file` it works ok.

shalvah:
Hmm, strange. Do examples in other languages show up?

MohammadMirsafaei:
sorry for late answering, yes it happens in other languages too

wyfiwyfi:
Thank you for your help with this. Very much appreciated.

PR #603, opened 2023-01-13
Fix nested filters when generating examples for PHP, JS and  Python

AdamiecRadek:
This PR changes the way of generating examples for languages `PHP`, `JS`, and `Python`
Previously there was an issue when you had nested filters defined for your route like
```
[
  'data' => 'filled|array',
  'data.*' => 'required|array',
  'data.*.is_file' => 'required|boolean',
]
```
Possible connected issue: https://github.com/knuckleswtf/scribe/issues/308 and https://github.com/knuckleswtf/scribe/issues/455

Now Scribe will nicely generate examples for them.

### DIFF
--- a/tests/Unit/WritingUtilsTest.php
+++ b/tests/Unit/WritingUtilsTest.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Knuckles\Scribe\Tests\Unit;
+
+use Knuckles\Scribe\Tests\BaseLaravelTest;
+use Knuckles\Scribe\Tools\WritingUtils;
+
+class WritingUtilsTest extends BaseLaravelTest
+{
+    /** @test */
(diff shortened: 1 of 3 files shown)


tests/Unit/WritingUtilsTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Knuckles\Scribe\Tests\Unit;
+
+use Knuckles\Scribe\Tests\BaseLaravelTest;
+use Knuckles\Scribe\Tools\WritingUtils;
+
+class WritingUtilsTest extends BaseLaravelTest
```
shalvah, line 8: No idea why I didn't have a test for this class already, so thanks for this. Now I need to add tests for the other things in that file.😅

shalvah (approved):

PR #605, opened 2023-01-19
added support for nested arrays as query parameters for bash examples

eschricker:
Fixed: When using nested arrays as query parameters, the bash examples throw an exception. 

Example: `filter[category][]=2&filter[category][ ]=6` broke the existing code. 

Solution: Check if the parameter is an array and handle it accordingly.


shalvah:
Can you add a test? You can follow the examples in https://github.com/knuckleswtf/scribe/pull/603

eschricker:
I restructured the code to support multiple nested arrays, as in the code for js, python and php. 

In addition, I have extended the referenced test.

PR #606, opened 2023-01-20
Add support for the Laravel Actions package

### LINKED ISSUE #604: Compatibility with the Laravel Actions package
We are using the Laravel Actions package (https://laravelactions.com/) in our application.

While Scribe properly discovers all the routes, it does not seem to pick-up any annotation we put in any of the places we have tried.

Basically, that helps avoiding to duplicate code or have boilerplate code when code could be used in multiple places (controller, job, command, etc.)

Route declaration is rather standard

```
Route::post('/auth/token')
    ->name("auth.token.store")
    ->uses(CreateProviderToken::class);
```

And the `CreateProviderToken` class is what serves as controller (but can also be turned easily in an artisan command):

```
class CreateProviderToken
{
    public function handle(string $apiKey): AccessToken
    {
        // [...] Logic to generate a token, e.g. with Sanctum
        return AccessToken::fromNewAccessToken($token);
    }

    //region Controller specific

    public function rules(): array
    {
        return [
            "api_key" => ["required", "string", Rule::exists('providers', 'api_key')],
        ];
    }

    public function asController(Request $request): JsonResponse
    {
        $token = $this->handle($request->string("api_key"));

        return response()->json($token);
    }

    //endregion
}
```

- `handle` method is what contains the logic to generate a token
- `asController` method is an entry point used when the action is called by a route
- `asCommand` method similarly would be the entry point when calling from Artisan

I have tried putting the following docblock both : before the `Route::post` declaration AND before the `asController` method, but it does not get read (title, content, tags are not in the generated docs)

I have also tried `artisan scribe:generate --force` (yaml have not been touched though).

How could I help Scribe to understand that I want to put my annotation on the asController method? I don't mind writing a plugin for that but I don't know where to start (I understood so far that plugins were more about adding tags, here my problem is to at least read the existing tags).

vpratfr:
docblock should be read from the asController method

fix #604

### DIFF
--- a/src/Tools/Utils.php
+++ b/src/Tools/Utils.php
@@ -105,7 +105,15 @@ class Utils
             if (is_array($uses)) {
                 return $uses;
             } elseif (is_string($uses)) {
-                return explode('@', $uses);
+                [$class, $method] = explode('@', $uses);
+                
+                // Support for the Laravel Actions package, docblock should be put on the asController method
+                if ($method === '__invoke' && method_exists($class, 'asController'))
+                {
+                    return [$class, 'asController'];
+                }
+
+                return [$class, $method];
             } elseif (static::isInvokableObject($uses)) {
                 return [$uses, '__invoke'];
             }


shalvah:
I'd prefer if there were also a test, but I'll merge this, since you're the one using it. 😜

Issue #551, opened 2022-11-08
Documenting Custom Rules


MissaelAnda:
- [x] I've read the [documentation](https://scribe.knuckles.wtf/laravel) and I can't find details on how to achieve this.

It would be cool if custom rules could be documented with description and example.

This:
```php
/**
 * @description This rule validates something
 * @example "This:Is:An:Example"
 */
class CustomRule implements Rule
{
    //
}
```

Produces:
![imagen](https://user-images.githubusercontent.com/31199350/200691715-457e384c-40bd-43e5-b317-f5639bf55966.png)

shalvah:
Good idea. PRs welcome.

I'd prefer a static `docs` method, though. It's simpler to implement and avoids us having to parse the annotation text, which can get tricky.

PR #611, opened 2023-02-04
Documentation for custom rules and closure rules

MissaelAnda:
## Document Custom Rule Classes and Closure Rules

Add a description to Closure rule with a DocComent:
```php
$request->validate([
    'complex_rule' => [
        /** This is a very specific validation therefore it needs a better description */
        function ($attribute, $value, $fail) { $fail('error'); }
    ],
]);
```

Add more info to a Custom Validation Rule:
```php
class CustomRule implements Rule
{
    // ...

    public static function docs(): array
    {
        return [
            'description' => 'This rule makes something very specific',
            'example' => '52KB',
        ];
    }
}
```

This is a followup for #551

I'm missing the tests and documentation changes, if you could point me where and how should I do them it'd be nice.

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -6,6 +6,7 @@ use Illuminate\Contracts\Validation\Rule;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Str;
+use Illuminate\Validation\ClosureValidationRule;
 use Knuckles\Scribe\Exceptions\CouldntProcessValidationRule;
 use Knuckles\Scribe\Exceptions\ProblemParsingValidationRules;
 use Knuckles\Scribe\Exceptions\ScribeException;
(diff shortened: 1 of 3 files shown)


shalvah:
The class that tests this is the `ValidationRuleParsingTest`. You should probably add two new test methods, something like `can_parse_custom_closure_rules` and `can_parse_custom_rule_classes`. You can follow the existing examples.

shalvah:
It feels a bit iffy to me to have an example defined on a rule. I guess it can work, but I'm uncertain if we want to go that path. For example, what happens if one parameter has this rule and others defined? The example will either be overwritten or hardcoded to this, which may or may not be unexpected behaviour. Maybe we should leave that out for now (after all, you can already specify examples for parameters in a FormRequest).

MissaelAnda:
The idea with the example is to define a default example in case none given like the string rules "setter"s, the problem maybe is that the example is always assigned. I'll change it to only set the example with the "setter" field and add the tests.

src/Extracting/ParsesValidationRules.php
```diff
@@ -164,300 +166,328 @@ protected function normaliseRules(array $rules): array
...
-            }
+                $dependentRules = ['between', 'max', 'min', 'size', 'gt', 'gte', 'lt', 'lte', 'before', 'after', 'before_or_equal', 'after_or_equal'];
+                if ($independentOnly && in_array($rule, $dependentRules)) {
+                    return false;
+                }
 
-            return true;
-        } catch (Throwable $e) {
-            throw CouldntProcessValidationRule::forParam($parameterData['name'], $rule, $e);
+                // Reminders:
```
shalvah, line 210: Btw, we can use early returns here to avoid this huge diff from indentation.

shalvah: Done.

shalvah:
Can you send in a docs PR as well?

MissaelAnda:
I just relized that I'm missing `InvokableRule`, it only needs to change `ParsesValidationRules` to import it and change line 194 to:
```php
if ($rule instanceof Rule || $rule instanceof InvokableRule) {
```

shalvah:
Added.

PR #616, opened 2023-02-15
Parses inline Enum Rules

MissaelAnda:
Adds support to parse inline Enum Rule class.


shalvah (commented):
Nice work! But don't forget that there's also the `Rule::enum()` version, which quite frankly, is what I'd prefer to use. See https://github.com/knuckleswtf/scribe/blob/6799496e3fa0e79cdcf48f68785c16a942dbc756/tests/Unit/ValidationRuleParsingTest.php#L519

shalvah (approved):
Nice, thank you!

PR #623, opened 2023-02-21
Fix sorting of multiple responses by status code

bram-pkg:
The `sortBy` function from Laravel does not modify the collection's sort order. It returns a new sorted collection.


shalvah:
Thanks! Could you add a test for this?

bram-pkg:
Yep, sure thing!

bram-pkg:
@shalvah I've added some unit tests!

shalvah (approved):

bram-pkg:
I've fixed the PHPStan error, my bad. It should be fine now.

PR #628, opened 2023-03-07
Ignore upgrade check if user hasn't published config yet

sangnguyenplus:
It will throw error if there is no config file in `config/scribe.php`.

We should check if this file exists or not, if it's not exists, just ignore upgrade config.

Before:
Screenshot 2023-03-07 at 15 42 57

After:
Screenshot 2023-03-07 at 15 45 05


shalvah:
Can you fix the tests?

sangnguyenplus:
@shalvah I have just fixed tests.

Please check again.

PR #639, opened 2023-03-27
Remove unnecessary randomisation from auth description

bram-pkg:
Whenever new documentation is generated, it should ideally not change if none of the backing endpoints or configuration has changed. I think this randomisation should not occur while generating the contents of API documentation.

### DIFF
--- a/src/Extracting/ApiDetails.php
+++ b/src/Extracting/ApiDetails.php
@@ -2,7 +2,6 @@
 
 namespace Knuckles\Scribe\Extracting;
 
-use Illuminate\Support\Arr;
 use Knuckles\Scribe\Tools\ConsoleOutputUtils;
 use Knuckles\Scribe\Tools\DocumentationConfig;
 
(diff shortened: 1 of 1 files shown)


shalvah:
This makes a lot of sense. My motivation for randomness is to vary things a bit between users, so not everyone's docs look exactly the same. Maybe I'll just use the documentation title as a seed. 🤔

bram-pkg:
I think it would make sense to have no randomisation and at all, it's good for libraries to be determinate about their output. Nowhere is it implied this is randomised. Ideally the whole auth section would be customisable, but removing the randomisation is the next best thing really.

shalvah:
I still favour a bit of variety, but I'll go with this now, especially as there are plans to introduce a translation system in future, which could allow users specify their own message.

bram-pkg:
Translations would indeed fix the core issue here, this is a good intermediary step though :)

PR #645, opened 2023-04-19
Update description to include Lumen

rin4ik:


shalvah:
Sorry, man. I'm probably not going to accept this. I'm actively trying to discourage Lumen usage, as it has become a third-class citizen even for the Laravel team itself. Besides, it's still obviously mentioned in a lot of places in our docs (unfortunately).

Issue #652, opened 2023-05-03
Generate document with `@apiResource` but without `@apiResourceModel`？


ccec1t91077:
### Scribe version

4.19.0

### Your question

Hi, I have some functional api, and I'm trying to use the tag `@apiResource` to generate the document, but the command `scribe:generate` will got an exception and failed if I use the tag `@apiResource` but without any `@apiResourceModel`.

In my case, I've some response(api resoruce) it doesn't need to display any data, we are using exception and error code to check if the result failed, it means I don't have any model or collection need to pass into api resource.

The resource will seems like：
```php
// I added default empty resource from here.
class SuccessResponse extends BaseResource
{
    public function __construct($resource = [])
    {
        parent::__construct($resource);
    }

    public function toArray(Request $request): array
    {
        return [];
    }
}
```

This allow my controller simple like：
```php
public function __invoke(Request $request)
{
    /*
        Do something functional ...
    */

    return new SuccessResponse();
}
```

Although I can define it by `@response`, but I wonder if you think this is a reasonable idea, or there does have ways I didn't notice, please let me know. 😃

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
Oof. it's not supported, so you'll have to do some other way (`@response`, `@responseFile`, custom annotation, ...). Or you can send in a PR.

ccec1t91077:
I've send a [PR](https://github.com/knuckleswtf/scribe/pull/662) for this little feature, and I'll send another PR to update the document after merged, thanks.

ccec1t91077:
Oops. Mistakenly clicked the close button.

ccec1t91077:
Close since merge, update document later.

PR #662, opened 2023-05-09
Support `@apiResource` with missing `@apiResourceModel`

ccec1t91077:
Update：

* Allow generate document without @apiResourceModel tag (but will remind you didn't use the @apiResourceModel tag)

This is a little feature to allow the document generate when the JsonResource is working with empty resource which means won't have any model to specify on the @apiResourceModel tag.

This might help some project using JsonResource to manage the api response which like webhook api or none data response api.

From [#652](https://github.com/knuckleswtf/scribe/issues/652).

### DIFF
--- a/src/Extracting/InstantiatesExampleModels.php
+++ b/src/Extracting/InstantiatesExampleModels.php
@@ -24,7 +24,7 @@ trait InstantiatesExampleModels
         array   $relations = [], ?ReflectionFunctionAbstract $transformationMethod = null
     )
     {
-        // Early return if JsonResource working with empty resource, there won't have an example model
+        // If the API Resource uses an empty resource, there won't be an example model
         if($type == null && $transformationMethod == null)
             return null;
 
--- a/src/Extracting/InstantiatesExampleModels.php
+++ b/src/Extracting/InstantiatesExampleModels.php
@@ -17,13 +17,17 @@ trait InstantiatesExampleModels
      * @param string[] $relations
      * @param \ReflectionFunctionAbstract|null $transformationMethod A method which has the model as its first parameter. Useful if the `$type` is empty.
      *
-     * @return \Illuminate\Database\Eloquent\Model|object
+     * @return \Illuminate\Database\Eloquent\Model|object|null
      */
     protected function instantiateExampleModel(
         ?string $type = null, array $factoryStates = [],
         array   $relations = [], ?ReflectionFunctionAbstract $transformationMethod = null
     )
     {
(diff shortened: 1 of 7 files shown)


shalvah:
Can you add a test for the expected behaviour?

ccec1t91077:
Hi, I've added tests but I'm not sure if I'm doing things correctly because I don't have much experience for writing tests, so if I did something bad please correct me, ty. 🙂

tests/Fixtures/TestEmptyApiResource.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->resource = $resource;
+    }
+
+    /**
+     * Get any additional data that should be returned with the resource array.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return array
+     */
+    public function with(Request $request)
```
shalvah, line 27:
```suggestion
    public function with($request)
```

src/Extracting/Shared/ApiResourceResponseTools.php
```diff
@@ -54,7 +54,9 @@ public static function getApiResourceOrCollectionInstance(
         array  $paginationStrategy = [], array $additionalData = []
     ): JsonResource
     {
-        $modelInstance = $modelInstantiator();
+        // If the JsonResource is working with empty $resource (such like an empty array), the $modelInstantiator will be null
+        // If $modelInstantiator is null set to an empty array as default
```
shalvah, line 58:
```suggestion
        // If the JsonResource is uses an empty $resource (e.g. an empty array), the $modelInstantiator will be null
        // See https://github.com/knuckleswtf/scribe/issues/652
```

src/Extracting/InstantiatesExampleModels.php
```diff
@@ -17,13 +17,17 @@ trait InstantiatesExampleModels
...
      *
-     * @return \Illuminate\Database\Eloquent\Model|object
+     * @return \Illuminate\Database\Eloquent\Model|object|null
      */
     protected function instantiateExampleModel(
         ?string $type = null, array $factoryStates = [],
         array   $relations = [], ?ReflectionFunctionAbstract $transformationMethod = null
     )
     {
+        // Early return if JsonResource working with empty resource, there won't have an example model
```
shalvah, line 27:
```suggestion
        // If the API Resource uses an empty resource, there won't be an example model
```

shalvah:
Nice, thanks!

PR #666, opened 2023-05-12
add JSON_PRETTY_PRINT for better body raw readable

yekta-kalantary:


shalvah:
I'm not sure about this yet. It was the case before, but I changed it (don't remember why, but I think it was also requested then).

PR #670, opened 2023-05-19
camelCase route parameter so multi-word model binding works

### LINKED ISSUE #669: Multi-word names not handled correctly when extracting route key from url param
### Scribe version

4.19.1

### PHP version

8.2.6

### Framework

Laravel

### Framework version

10.11.0

### Scribe config

```ruby
type => "laravel"
```

### What happened?

After running `php artisan scribe:generate` the documentation has wrong url parameter for multi-word paths e.g. `/api/user-address/{user_address}`. Url parameter defaults to `id` but I'm having `uuid` returned from the method `getRouteKeyName()`. I tracked the issue down to `getRouteKeyFromModel(...)` function. It seems the `$paramName` is `user-address` while the array `$typeHintedEloquentModels` has a key named `userAddress` so it can't find it.

https://github.com/knuckleswtf/scribe/blob/3cd295f6b7d235f47350ab6b92c3e4571bca20f5/src/Extracting/Shared/UrlParamsNormalizer.php#L186-L194

### Docs

Rasmus-Bertell:
Changes multi-word url parameters from snake_case to camelCase so it matches the model name in array. Propably should handle the case changing in some other way, I'll look into it later.

I don't know how to write a test for this, so any help is welcome.

closes #669


shalvah:
Sorry, haven't had a chance to try this and decide if it's what we actually want. Will probably not be able to check it out until end of the month, when I'm back from vacation.

Rasmus-Bertell:
> Sorry, haven't had a chance to try this and decide if it's what we actually want. Will probably not be able to check it out until end of the month, when I'm back from vacation.

I still need to figure out how to write the test case for this, it would demonstrate the issue better in my opinion.

shalvah:
Maybe take a look at the existing tests. I can't recall off the top of my head, but there's a class where we test some different URL parameter matching scenarios.

src/Extracting/Shared/UrlParamsNormalizer.php
```diff
@@ -185,6 +185,9 @@ protected static function getInlineRouteKey(Route $route, string $paramName): ?s
      */
     protected static function getRouteKeyFromModel(string $paramName, array $typeHintedEloquentModels): ?string
     {
+        // Ensure param name is in camelCase so it matches the model name
```
shalvah, line 188:
```suggestion
        // Ensure param name is in camelCase so it matches the argument name (e.g. The '$userAddress' in `function show(BigThing $userAddress`)
```

Issue #671, opened 2023-05-26
Localization is not working


ccec1t91077:
### Scribe version

4.19.1

### PHP version

8.2.5

### Framework

Laravel

### Framework version

10.12.0

### Scribe config

```ruby
------ SAME AS DEFAULT CONFIG ------
```

### What happened?

Localization is not working, translations always using json from `/vendor/knuckleswtf/scribe/lang/en.json`.

### Steps To Reproduce

1. `create project laravel/laravel test-project`
2. Setup Laravel `.env`
3. `composer require --dev knuckleswtf/scribe`
4. `php artisan vendor:publish --tag=scribe-translations`
5. Modify the `/lang/vendor/scribe/en.json` make some changes of defaults
6. `php artisan scribe:generate`
7. Check the doc it generated, **not using the `/lang/vendor/scribe/en.json` to generate**
8. Move the `lang` folder inside `resources`
9. `php artisan scribe:generate`
10. Check the doc it generated, **not using the `/resources/lang/vendor/scribe/en.json` to generate**

I met #657 this week. After upgrade to 4.19.1 this issue shows up.

Seems like it won't try to load localization from the published path.

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
This whole thing is a mess. `.php` translation files and `.json` translation files have different behaviours, and interact with your app differently. Laravel 10 also changed the default location of translation files, and the publish path as well.

shalvah:
Fixed in #673

PR #672, opened 2023-05-26
Fix the bug not using published localization when generating

ccec1t91077:
Fix: #671


shalvah:
Unfortunately, this doesn't work on Laravel 8. I think I've found a workaround, though..

shalvah:
Okay, I think I've got it. Can you test with the `localization` branch? See this PR for instructions: https://github.com/knuckleswtf/scribe/pull/673

shalvah:
Gonna go ahead and close this.

PR #673, opened 2023-05-27
Implement custom translation layer

shalvah:
Translations should work correctly onLaravel 8,9 and 10 now. 

To publish translations, run `php artisan vendor:publish --tag="scribe-translations"`. This will create the file  `resources/lang/scribe.php` (or `lang/scribe.php`).

This may be a bit of a minor breaking change to the previous translation system—they will be ignored. I would recommend deleting any previously published Scribe translations (the `en.json` file).

The new system is:
- PHP-based, not JSON-based. Easier to manage.
- independent of locale. Just edit the published file as you like. If you want to use Scribe's default for certain strings, you can just delete them from your published file.
- (TO-DO) safer. If a translation string is not found, it will raise an error, rather than print the raw `scribe::` string. That way users don't get unexpected text in their docs.


ccec1t91077:
I've tested `localization` branch in Laravel 8 and 10 clean project, works fine. :D

### Testing process

1. `composer require --dev knuckleswtf/scribe:dev-localization`
2. `php artisan vendor:publish --tag=scribe-config`
3. Setup .env
4. `php artisan scribe:generate`
5. Check docs
6. `php artisan vendor:publish --tag="scribe-translations"`
7. Modify `lang/scribe.php`
8. `php artisan scribe:generate`
9. Check docs

shalvah:
Nice, thanks!

PR #680, opened 2023-06-06
Provide ability to extend Header attribute

cosmastech:
Our application has an extra header that needs to be sent with most (but not all) requests. It is not ideal to have to write `#[Header('Special-Header', 'this-needs-to-be-created-dynamically-at-runtime')]`

This PR makes it possible to extend the Header attribute.


shalvah:
Nice, good to see some cool use cases for attributes.

Issue #688, opened 2023-06-20
Should stop add full stops on custom description


ccec1t91077:
### Scribe version

4.21.1

### Your question

The description support markdown format, so I think if you want it, it more easier to add yourself, but if you don't need the full stops, there will be a little issue on your document.

For example the situation what I met:

I'm trying to wrap the line of `Example:` after description, and also add a line `Default: 1`.

### My code in request
```php
/**
 * For scribe document.
 *
 * @return array
 */
public function queryParameters()
{
    return [
        'page' => [
            'description' => 'Current page.<br> Default: 1<br>'
        ]
    ];
}
```

### The output document
![image](https://raw.githubusercontent.com/ccec1t91077/Storage/master/image.png)

You can see the full stops it added after description, but in my case, I'm trying to wrap line, the full stops will appear at the place I do not expect.

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

ccec1t91077:
In some of case, it will auto added a full stops after your parameter description, for example if you have a custom validation rule and there's a rule description for scribe.

If someone want to wrap line and face same issue as here, quick solution is add the wrap line mark to your custom rule description instead of parameter description.

PR #689, opened 2023-06-20
remove the full stops as default it added

ccec1t91077:
Try to solve #688.

Updated:
* Remove the full stops it added as default on description.


shalvah:
Hey, I'll merge this if you can fix the tests. Shouldn't be a big deal; just update some tests (and probably remove a few).

ccec1t91077:
Sorry, I got things to work on my job. 

For now, I already comfirm that it might cause some of use case document broke (terrible readable).

Because I didn't confirm it affect by document of each situation, so I'll close it at first, until when I have time to finish it.

PR #691, opened 2023-06-21
Also show empty array query parameters in Postman

laudaikinhdi:
When query param empty I want to show Postman

image


src/Writing/PostmanCollectionWriter.php
```diff
@@ -300,6 +300,15 @@ protected function generateUrlObject(OutputEndpointData $endpointData): array
                         'disabled' => !$parameterData->required && empty($parameterData->example),
                     ];
                 }
+                if (empty($values)) {
```
shalvah, line 303:
```suggestion
                // If there are no values, add one entry so the parameter shows up in the Postman UI.
                if (empty($values)) {
```

PR #712, opened 2023-08-17
add empty check for properties

rpmccarter:
Objects with empty properties are being serialized as `[]` in the resulting `openapi.yaml`, which is invalid OpenAPI. This was already being handled in a couple places, just missed a few. I also replaced the `$EMPTY` class prop with a function that also performs the empty check. 

Luckily, `TheSideProjectAPI` actually has an example of this bug! I tested with that repo and the fix seemed to work. 

First time working in php, pointers welcome!

### DIFF
--- a/src/Writing/OpenAPISpecWriter.php
+++ b/src/Writing/OpenAPISpecWriter.php
@@ -233,6 +233,7 @@ class OpenAPISpecWriter
                 $schema['properties'][$name] = $fieldData;
             }
 
+            // We remove 'properties' if the request body is an array, so we need to check if it's still there
             if (array_key_exists('properties', $schema)) {
                 $schema['properties'] = $this->objectIfEmpty($schema['properties']);
             }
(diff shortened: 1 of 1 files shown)


src/Writing/OpenAPISpecWriter.php
```diff
@@ -241,6 +233,9 @@ protected function generateEndpointRequestBodySpec(OutputEndpointData $endpoint)
                 $schema['properties'][$name] = $fieldData;
             }
 
+            if (array_key_exists('properties', $schema)) {
```
rpmccarter, line 236: `'properties'` gets erased when an array is encountered, so we need to check if it's still there

shalvah, line 236: Let's put that in the code:

```suggestion
            // We remove 'properties' if the request body is an array, so we need to check if it's still there
            if (array_key_exists('properties', $schema)) {
```

shalvah (approved):
Nice! The function solution is indeed cleaner!

PR #713, opened 2023-08-20
Support limitng parameter values (enums)

### LINKED ISSUE #626: enum support as parameters
- [x] Version info - I'm using `scribe 4.16.0` with `laravel 10`
- [x] I've read the [documentation](https://scribe.knuckles.wtf/laravel) and I can't find details on how to achieve this.
- [x] I've searched the word `enum` in the documentation, but only found one occurence which was irrelevant.

The question is rather related to the generated `openapi.yaml` file.   
As shown in the [documentation](https://swagger.io/docs/specification/adding-examples/) page, openapi specification supports `enum`s as parameters. 
```
parameters:
  - in: query
    name: status
    schema:
      type: string
      enum: [approved, pending, closed, new]
      example: approved 
```   
This helps the documentation to show dropdown menus for parameters having several predefined values, as in the following image:   

image
  
**Question**
So, do we have a way to annotate a parameter in the comment blocks so that scribe generates the `openapi.yaml` file with `enum`  value list? I guess I can override this via `nn.yaml` or `config/scribe.php` files, but it would be nice to do this via comment blocks.  
Thanks!

shalvah:
Adds support for limiting parameter values. Works with either a PHP 8.1 enum or an array of values.

```php
#[BodyParam("status", enum: ['approved', 'pending', 'closed', 'new'])]
#[BodyParam("status", enum: ArticleStatus::class)]
```

Also supported in docblocks:

```
@bodyParam status string Enum: approved, pending, closed, new
@bodyParam status string Enum: approved, pending, closed, new. Example: new
```

This will:
1. ensure the generated example is from one of these.
2. add a line in the docs that says "Must be one of...", and add these values
3. set the `enum` key in the OpenAPI spec

It does not currently restrict the values that can be entered in Try It Out, as there may still be some value in seeing how invalid values are responded to.

This may be a breaking change, as it replaces the earlier manual description setting (automatically generated for validation rules) with a structured field. 

In the internal Scribe API, this adds a new `enumValues` field to the parameter types. This is an array of all valid values.

Closes #626


rabol:
This is a breaking change

After I updated to the latest version i get this error:

```
 TypeError 

  Cannot assign null to property Knuckles\Camel\Extraction\Parameter::$enumValues of type array

  at vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php:48
```

PR #714, opened 2023-08-21
Exclude Authorization header from openapi specs.

janolivermr:
In addition to the existing 2 headers, the Authorization header is also not allowed. Instead it is taken care of by the security schemes


shalvah:
Thanks! And good call on the lowercasing too.

PR #720, opened 2023-09-02
Fix sample body with array of object

stephane-monnot:
Hi,

Fix sample body with an object of array. For example :

```php
    #[BodyParam('fields', 'object[]', 'Form\'s fields', required: false)]
    #[BodyParam('fields[].name', 'string', 'Field\'s name', required: true, example: 'experience')]
    #[BodyParam('fields[].label', 'string', 'Field\'s label', required: true, example: 'Expérience')]
    #[BodyParam('fields[].type', 'string', 'Field\'s type', required: true, example: 'textarea')]
    #[BodyParam('fields[].order', 'number', 'Field\'s order', required: true, example: 1)]
```

Before
```json
{
    "name": "Experience Form",
    "fields": {
        "name": "experience",
        "label": "Exp\u00e9rience",
        "type": "textarea",
        "order": 1
    }
}
```

After
```json
{
    "name": "Experience Form",
    "fields": [
        {
            "name": "experience",
            "label": "Exp\u00e9rience",
            "type": "textarea",
            "order": 1
        }
    ]
}
```

### DIFF
--- a/src/Tools/WritingUtils.php
+++ b/src/Tools/WritingUtils.php
@@ -110,7 +110,7 @@ class WritingUtils
         $output = isset($braces[0]) ? "{$braces[0]}\n" : '';
         foreach ($cleanQueryParams as $parameter => $value) {
             $output .= self::printSingleQueryParamAsKeyValue($value, $spacesIndentation, $startLinesWith, $quote,
-                                                             $parameter, $delimiter, $endLinesWith);
+                $parameter, $delimiter, $endLinesWith);
         }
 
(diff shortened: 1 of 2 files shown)


shalvah:
Cool. Could you rename some of those test methods? `get_sample_body()` and `bodyParams()` are too generic names for a test that's designed to test certain specific things (in this case, array fields).

stephane-monnot:
Thank you for the review. I have renamed the methods.

Issue #721, opened 2023-09-07
Approach to multi-tenancy docs


onovaes:
### Scribe version

4.23.1 

### Your question

I'm making a multitenant application with package (/spatie/laravel-multitenancy) and generating static documentation.

When I access /docs/ the Base URL always get the APP_URL. 

How do I generate each /docs/ with Base URL correct for each tenant?

What is the best approach to multi-tenancy?

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
Hmm, I don't have an answer out of the box for you. First thing is that you should definitely be using `laravel` type, because that's the only way you can add some dynamic behaviour to your docs.

I'd advise you to read the docs about the URLs Scribe uses (displaying vs Try It Out vs response calls), if you haven't already done so.

My guess is that you can use `laravel` type and customise the generated Blade views to display a dynamic base URL instead, or something like that. Or you could even set `base_url` to a dynamic string, like `'config ("app.tenant_url")`. It _might_ work.

shalvah:
> Or you could even set base_url to a dynamic string, like 'config ("app.tenant_url"). It might work

Okay, tested it and it doesn't work. But pushed a patch to master and it sort of does, now. Can you test that out?

pageantry:
you can try [this approach](https://github.com/knuckleswtf/scribe/issues/100#issuecomment-2907805503)

Issue #717, opened 2023-08-25
Add more integraion tests


shalvah:
As Scribe has gotten more features (attributes, inline validators, etc), it's become more difficult to test them in combination. We only have unit tests. We should endeavour to add more end-to-end tests. Will help avoid issues like #715

PR #723, opened 2023-09-07
Support dynamic base URL

shalvah:
- makes the workaround in #721 work
- still needs tests (linking #717)

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -17,7 +17,8 @@ return [
     'description' => '',
 
     /*
-     * The base URL displayed in the docs. If this is empty, Scribe will use the value of config('app.url').
+     * The base URL displayed in the docs. If this is empty, Scribe will use the value of config('app.url') at generation time.
+     * If you're using `laravel`` type, you can set this to a dynamic string, like '{{ config("app.tenant_url") }}' to get a dynamic base URL.
      */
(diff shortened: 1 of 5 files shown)


deefour:
@shalvah I haven't had the time to trace `$baseUrl` back further yet, but I wanted to point out with the 'base_url' left as `null` for both scribe proper and 'try it out,' pulling in this change breaks 'try it out.'

We end up with `undefined/our/endopint/path/here` seemingly because scribe isn't setting up `$baseUrl` properly for the views.

image

deefour:
@shalvah I think I sorted it, but wish to communicate in case it helps for later changelogs.

`baseUrl` in the scribe default `index.blade.php` was changed/replaced/... with `tryItOutBaseUrl` at some point. Our upgrade from 4.18 to 4.25 became broken because of this, and it seems it wasn't documented in the Changelog (did I miss it)?

shalvah:
Ah, yeah, that's from the localisation feature released in 4.25.0. https://github.com/knuckleswtf/scribe/pull/647

Sorry about that. Tbh, it's not so easy to maintain backwards compatibility for custom themes, but I'll keep that in mind going forward. Maybe I'll just add both old and new variables instead of renaming.

PR #725, opened 2023-09-13
missing authorization headers issue fixed

satputekuldip:
I was not able to test the try it out feature because of the headers... Few lines was missing in upgraded version...

### DIFF
--- a/resources/js/tryitout.js
+++ b/resources/js/tryitout.js
@@ -231,8 +231,12 @@ async function executeTryOut(endpointId, form) {
     const urlParameters = form.querySelectorAll('input[data-component=url]');
     urlParameters.forEach(el => (path = path.replace(new RegExp(`\\{${el.name}\\??}`), el.value)));
 
-    const headers = Object.fromEntries(Array.from(form.querySelectorAll('input[data-component=header]'))
-        .map(el => [el.name, el.value]));
+    const headers = JSON.parse(form.dataset.headers);
+    // Check for auth param that might go in header
+    if (form.dataset.authed === "1") {
+        const authHeaderEl = form.querySelector('input[data-component=header]');
+        if (authHeaderEl) headers[authHeaderEl.name] = authHeaderEl.dataset.prefix + authHeaderEl.value;
+    }
 
     // When using FormData, the browser sets the correct content-type + boundary
     let method = form.dataset.method;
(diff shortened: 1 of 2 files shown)


resources/js/tryitout.js
```diff
@@ -231,8 +231,12 @@ async function executeTryOut(endpointId, form) {
...
     urlParameters.forEach(el => (path = path.replace(new RegExp(`\\{${el.name}\\??}`), el.value)));
 
-    const headers = Object.fromEntries(Array.from(form.querySelectorAll('input[data-component=header]'))
-        .map(el => [el.name, el.value]));
+    const headers = JSON.parse(form.dataset.headers);
+    // Check for auth param that might go in header
+    if (form.dataset.authed === "1") {
+        const authHeaderEl = form.querySelector('input[data-component=header]');
+        if (authHeaderEl) headers[authHeaderEl.name] = authHeaderEl.dataset.prefix + authHeaderEl.value;
+    }
```
shalvah, line 239: I don't think this PR is right. Where does `form.dataset.headers` come from? There's no data-headers attribute on the form. And how do we know the auth header is the first item matching 'input[data-component=header]'?

satputekuldip: It comes from config 

[config/scribe.php Line no 76](https://github.com/knuckleswtf/scribe/blob/b32959d441d425c3c1616bf0f8b7e26524ac3883/config/scribe.php#L76)

and output is like

```html
<form id="form-POSTapi-login" data-method="POST"
      data-path="api/login"
      data-authed="0"
      data-hasfiles="0"
      data-isarraybody="0"
      data-headers='{"Content-Type":"application\/json","Accept":"application\/json"}'
      autocomplete="off"
      onsubmit="event.preventDefault(); executeTryOut('POSTapi-login', this);">
```

satputekuldip: and also this was in older version but removed in newer version don't know why...

satputekuldip: https://github.com/knuckleswtf/scribe/blob/028d6064a593ac119d94f53414fbecb871670b2e/resources/js/tryitout.js#L230

satputekuldip: I've added missing data-headers='@json($endpoint->headers)' parameter

shalvah: Dude, you can't just add back something that was removed without trying to find out why.  This is _not_ good work. I'm sorry, I'll be closing this. You've made me waste time investigating this, only to find that you just copied an older version of these files. WTF. I suspect you're on an older version of Scribe and trying to change something in a newer version.

I'm closing this PR. If you wish to try again, please include a proper explanation of the issue you faced (with screenshots of before and after), and then explain how your fix actually changes things, not just copy/paste from an older version.

PR #726, opened 2023-09-13
* New config option to allow more customisation on the servers added …

ProjectZero4:
…to the open api yaml

See the below screenshots to see how the generated yaml changes

**OLD YAML**

image

**NEW YAML**

image

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -23,6 +23,16 @@ return [
     'base_url' => null,
 
     /*
+     * An array of servers available for the generated docs.
+     * Overrides base_url when generating the servers list for the openapi.yaml file.
+     */
+    'servers' => [
+//        [
+//            'url' => null,
+//            'description' => null,
+//        ],
+    ],
+    /*
      * Tell Scribe what routes to generate documentation for.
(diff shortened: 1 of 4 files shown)


ProjectZero4:
Just to explain the use case as I've just realised that the generated html and blade file that doesn't take advantage of this change.

We currently use a different provider for our api doc view but would like to use this package to start automating our generation of our yaml file

shalvah:
Sorry, gonna also close this, for the same reason as #729 

Also, small note: I always find it a bit funny how new contributors are quick to suggest a "new config item". There is such a thing as too many config items—it harms usability for users, increases maintenance overhead for maintainers (not to mention possibly multiplying the number of code paths). Every one wants to add a new config item for their use case, but if I accepted all these, and you as a new user installed this package and opened the config file to see 200 options, that would put you off.

So (even apart from the OpenAPI thing), my default reaction to anything proposing a new config item is No. There has to be a strong use case for it, a possibility that it will be valuable to more than just a few people, or the impossibility of achieving it any other way. Just letting you know. I'd say that, if you're thinking of adding a new config item, it's best to raise it as an issue for discussion first.

PR #727, opened 2023-09-14
* auth.extra_info now propagates through to the security schema as de…

ProjectZero4:
…scription

Please see the screenshots below of the openapi.yaml file

**Before**
image

**After**

image


shalvah:
Nice!

PR #728, opened 2023-09-14
 * When not using dot notation, overrides no longer remove all the ch…

ProjectZero4:
…ild keys in any parent array key. Instead, they are merged together.

**Note:** Dot notiation is still supported and I've also added support for the postman overrides too

Below i've put a screenshot of the overrides config and a before and after of the openapi.yaml file

**OVERRIDES**
image

**BEFORE**
image

**AFTER**

image

### DIFF
--- a/src/Writing/Writer.php
+++ b/src/Writing/Writer.php
@@ -2,6 +2,7 @@
 
 namespace Knuckles\Scribe\Writing;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Storage;
 use Knuckles\Scribe\Tools\ConsoleOutputUtils as c;
 use Knuckles\Scribe\Tools\DocumentationConfig;
@@ -127,7 +128,7 @@ class Writer
         $writer = app()->makeWith(PostmanCollectionWriter::class, ['config' => $this->config]);
 
         $collection = $writer->generatePostmanCollection($groupedEndpoints);
-        $overrides = $this->config->get('postman.overrides', []);
+        $overrides = Arr::dot($this->config->get('postman.overrides', []));
         if (count($overrides)) {
             foreach ($overrides as $key => $value) {
                 data_set($collection, $key, $value);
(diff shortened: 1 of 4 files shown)


shalvah:
That's fair, but I'm wondering what happens if you _do_ want to replace it. After all, you can still achieve this by listing the keys manually, but if I accept this, you can no longer replace.

ProjectZero4:
The main thing that got me looking into this is that by opting into any overrides in this manner you immediately lose anything that was done by the package like the `title` and `description. 

Is there actually a use case for removing keys beyond setting them to `null`? i.e if you wanted to remove `title` you can do 

```php
'overrides' => [
'info.title' => null
]
```

**OR**

```php
'overrides' => [
    'info' =>[
        'title' => null,
        // other keys ...
    ]
]
```

I would expect the main use case for this particular feature would be to add as opposed to remove

What do you think?

shalvah:
> by opting into any overrides in this manner you immediately lose anything that was done by the package

Are you sure? The code uses `data_set`. Specifying `info.version` should set the `version` key on the `info` object only. You shouldn't lose any other `info` fields.

ProjectZero4:
Yeah to be explicit 

On master if you had something like:

```php
'overrides' => [
    'info' => [
        'version' => '2.4.1',
    ],
],
```
Then you will lose `title` and `description` on the generated yaml. 

On master if you did the following:

```php
'overrides' => [
    'info.version' => '2.4.1',
],

```
You would keep `title` and `description`


While I understand what you mean by:
>  you can still achieve this by listing the keys manually, but if I accept this, you can no longer replace.

The main question I think to answer is if the default behaviour should be to remove everything when using nested arrays or if it should only change what's specified in the config.

shalvah:
That's all fine and good, but still, what's the point? Like, this works as expected:

```php
'overrides' => [
    'info.version' => '2.4.1',
    'info.contact' => [
        'name' => '',
    ],
],
```

I believe in convenience, but I don't think being able to do

```php
'overrides' => [
    'info' => [
        'version' => '2.4.1',
        'contact' => [
            'name' => '',
        ],
],
```

is such a step up.

Are there specific use cases I'm missing?

ProjectZero4:
Just done some research on Laravel's `mergeConfigFrom` and that behaves the same way as `master`. Happy to close since this change would break that consistency with laravel

PR #737, opened 2023-09-27
Adding Support for UUID Type

Jiraiya43:
This pull request introduces support for UUID (Universally Unique Identifier) types within your application.

### DIFF
--- a/src/Extracting/ParamHelpers.php
+++ b/src/Extracting/ParamHelpers.php
@@ -95,6 +95,7 @@ trait ParamHelpers
                 return $max ? $faker->numberBetween((int)$min, (int)$max) : $faker->randomFloat();
             },
             'boolean' => fn() => $faker->boolean(),
+            'uuid' => fn() => $faker->uuid(),
             'string' => fn() => $size ? $faker->lexify(str_repeat("?", $size)) : $faker->word(),
             'object' => fn() => [],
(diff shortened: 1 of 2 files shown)


shalvah:
I'm not sold on this. We try to restrict ourselves to the most common primitive data types, not subtypes. If we add a `uuid` type, what's stopping us from adding an `ip` type, or an `email` type?

Jiraiya43:
In fact, when I submitted the PR, I had the same thought. However, I thought that rather than restrict the options, we could use Faker as a generator and stick to the types it offers. What particularly frustrated me was that when I renamed the field to "uuid", it generated an example UUID, but when I set the type to "uuid", an example"string" type was provided instead.

shalvah:
That's an interesting proposal. Will need to consider it some more.

medeirosinacio (approved):

shalvah:
> In fact, when I submitted the PR, I had the same thought. However, I thought that rather than restrict the options, we could use Faker as a generator and stick to the types it offers. What particularly frustrated me was that when I renamed the field to "uuid", it generated an example UUID, but when I set the type to "uuid", an example"string" type was provided instead.

I know it's been a while, but this is a good idea, and I could get behind it if you want to implement it. Esssentially:
- you can use any type provided by Faker
- Scribe will use that as a hint to generate the correct example

Closing this PR and making an issue from that.

PR #742, opened 2023-10-08
GET method query param set from request

sudippalash:


sudip-fe:
> Please tell me what this is supposed to do, and include a before/after if you want me to merge it.

Your package is awesome if developer is using DocBlocks in every function. But when developer not use DocBlocks before his function then your package generate from validation roles. It's also awesome but **generate from validation roles** only applicable for POST & PUT route. GET route query param not working for this scenario, That's why i am adding this code. 

Check the below image:
image

shalvah:
Ah, I see. But that isn't the right solution. Check the docs for how to use validation rules for query parameters. https://scribe.knuckles.wtf/laravel/documenting/query-body-parameters#using-validation-rules-for-query-parameters

PR #749, opened 2023-10-16
Fixed parsing of nested fields in validation rules

raphaelflash:
## Description
If there are nested fields in the request validation rules, the parsing is incorrect.
This PR fixes this, so that it also works when an array is expected, see examples below.

### Screenshots
Before the change
Bildschirmfoto 2023-10-16 um 03 12 22

After the change
Bildschirmfoto 2023-10-16 um 03 11 48

## Example request class
```php
class ExampleRequest extends FormRequest
{
    public function rules()
    {
        return [
            '*.foo' => 'required|array',
            '*.foo.*' => 'required|array',
            '*.foo.*.bar' => 'required',
        ];
    }

    public function bodyParameters()
    {
        return [
            '*.foo' => [
                'description' => 'This is not shown (intended?).',
            ],
            '*.foo.*' => [
                'description' => 'A nested field.',
            ],
            '*.foo.*.bar' => [
                'description' => 'A double-nested field.',
            ],
        ];
    }
}
```
**Note:** The first description (of `*.foo` is ignored and will not be shown by scribe. I'm not sure if this is intended?

## Example request body
Single object example
```json
[
  {
    "foo": [
      {
        "bar": "hello world"
      }
    ]
  }
]
```

Multi object example
```json
[
  {
    "foo": [
      {
        "bar": "hello world"
      }
    ]
  },
  {
    "foo": [
      {
        "bar": "hallo welt"
      },
      {
        "bar": "hej världen"
      }
    ]
  }
]
```

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -145,9 +145,14 @@ trait ParsesValidationRules
         // Now this will return the complete ruleset.
         // Nested array parameters will be present, with '*' replaced by '0'
         $newRules = Validator::make($testData, $rules)->getRules();
-
-        // Transform the key names back from 'ids.0' to 'ids.*'
+       
         return collect($newRules)->mapWithKeys(function ($val, $paramName) use ($rules) {
+            // Transform the key names back from '__asterisk__' to '*'
+            if (Str::contains($paramName, '__asterisk__')) {
+                $paramName = str_replace('__asterisk__', '*', $paramName);
+            }
+
+            // Transform the key names back from 'ids.0' to 'ids.*'
             if (Str::contains($paramName, '.0')) {
                 $genericArrayKeyName = str_replace('.0', '.*', $paramName);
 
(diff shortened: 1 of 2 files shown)


src/Extracting/ParsesValidationRules.php
```diff
@@ -145,9 +145,14 @@ protected function normaliseRules(array $rules): array
         // Now this will return the complete ruleset.
         // Nested array parameters will be present, with '*' replaced by '0'
         $newRules = Validator::make($testData, $rules)->getRules();
-
-        // Transform the key names back from 'ids.0' to 'ids.*'
+       
         return collect($newRules)->mapWithKeys(function ($val, $paramName) use ($rules) {
+            // Transform the key names back from '__asterisk__' to '*'
+            if (Str::contains($paramName, '__asterisk__')) {
```
shalvah, line 151: Hi, where does the `__asterisk__` come from? I can't find that anywhere in the codebase.

shalvah: Ah, never mind; I see, it's from the framework.

shalvah (approved):
Thanks!

PR #755, opened 2023-11-01
Support specifying `Example: null` in annotations

liverocaisson:
Allow null parameter in the generation of API example specification

```
@bodyParam col_id integer required ID Example: null 
@bodyParam col_name string name Example: null
```

Example of use 
We want to put null value in the database.
We want to pass null as ID to create a new ID. //presence of parameter required
We want to pass null instead of '' string.

### DIFF
--- a/camel/Extraction/Parameter.php
+++ b/camel/Extraction/Parameter.php
@@ -13,7 +13,7 @@ class Parameter extends BaseDTO
     public mixed $example = null;
     public string $type = 'string';
     public array $enumValues = [];
-    public bool $hasExample = false;
+    public bool $exampleWasSpecified = false;
 
     public function __construct(array $parameters = [])
     {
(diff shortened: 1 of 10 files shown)


liverocaisson:
`
@bodyParam col_name string name Example: null
`
The null value is written in specification. Seems that if we don't put required, this parameter is ignored.
How to not ignore intentional null example, non-required ?

Can you give me some guidance about the file to modify.

liverocaisson:
I have managed to found why intentional null example are ignored

I have added a if condition to not ignore null example if the example are intentionally written in the comment as the example below

`@bodyParam col_name string name Example: null`

[src/Extracting/Extractor.php](https://github.com/knuckleswtf/scribe/pull/755/commits/a085e4f8532025f953253ede408efabdeb42d388#diff-37db58f7768fd51335e3fdef0edf016b9402566bbbfe50030cf8a74e8febf3d9)

shalvah:
Thanks!

PR #760, opened 2023-11-07
Allow config and temp directories to be configured independently.

ArclightHub:
### Whats changed?

- **Added** the ability to configure the `scribe.php` (or custom php config file) independently of the temp directory.
- **Added** a `docker-compose` setup to allow others to quickly run the test suite on their local machines without needing to install the test stack onto their environment.

### How do I run the tests in the new docker container?
- Simply run `clear && docker-compose up -d && docker logs -f scribe_app_1` and the tests are run and results printed into your terminal!

### DIFF
--- a/src/Extracting/ApiDetails.php
+++ b/src/Extracting/ApiDetails.php
@@ -2,6 +2,7 @@
 
 namespace Knuckles\Scribe\Extracting;
 
+use Knuckles\Scribe\Configuration\CacheConfiguration;
 use Knuckles\Scribe\Tools\ConsoleOutputUtils as c;
 use Knuckles\Scribe\Tools\Utils as u;
 use Knuckles\Scribe\Tools\DocumentationConfig;
@@ -23,9 +24,9 @@ class ApiDetails
 
     private array $lastKnownFileContentHashes = [];
 
-    public function __construct(DocumentationConfig $config = null, bool $preserveUserChanges = true, string $docsName = 'scribe')
+    public function __construct(DocumentationConfig $config = null, bool $preserveUserChanges = true, CacheConfiguration $docsName)
     {
-        $this->markdownOutputPath = ".{$docsName}"; //.scribe by default
+        $this->markdownOutputPath = $docsName; //.scribe by default
         // If no config is injected, pull from global. Makes testing easier.
         $this->config = $config ?: new DocumentationConfig(config($docsName));
         $this->baseUrl = $this->config->get('base_url') ?? config('app.url');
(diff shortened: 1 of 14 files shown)


src/Extracting/ApiDetails.php
```diff
@@ -23,9 +24,9 @@ class ApiDetails
 
     private array $lastKnownFileContentHashes = [];
 
-    public function __construct(DocumentationConfig $config = null, bool $preserveUserChanges = true, string $docsName = 'scribe')
+    public function __construct(DocumentationConfig $config = null, bool $preserveUserChanges = true, CacheConfiguration $docsName)
```
shalvah, line 27: Can you fix this?

src/GroupedEndpoints/GroupedEndpointsFactory.php
```diff
@@ -24,12 +28,12 @@ public static function fromApp(
         GenerateDocumentation $command,
         RouteMatcherInterface $routeMatcher,
         bool $preserveUserChanges,
-        string $docsName = 'scribe'
+        CacheConfiguration $docsName
```
shalvah, line 31: This seems like mismatching parameter names? Name `$docsName` and type `CacheConfiguration` (also in the other places in this file)
```

File: src/Writing/Writer.php [hunk 3]
```diff
@@ -86,8 +87,8 @@ protected function writePostmanCollection(array $groups): void
                 $collectionPath = "{$this->staticTypeOutputPath}/collection.json";
                 file_put_contents($collectionPath, $collection);
             } else {
-                Storage::disk('local')->put("{$this->docsName}/collection.json", $collection);
-                $collectionPath = Storage::disk('local')->path("$this->docsName/collection.json");
+                Storage::disk('local')->put($this->docsName->getScribeConfigFile() . "/collection.json", $collection);
+                $collectionPath = Storage::disk('local')->path($this->docsName->getScribeConfigFile() . "/collection.json");
```
shalvah, line 91: If we're doing this, perhaps we should move away from concatenation, and just make it a resolver (a la Laravel's `storage_path(...)`)? Somethig like:

```php
Storage::disk('local')->path($this->paths->outputFile("/collection.json"));
```

shalvah (changes requested):
I like the goal here, but I think the API can be improved. I thnk there are quite a few naming mismatches, which make it harder to follow (`$configObject` implies this is the general Scribe config, but it's a very limited subset of the configuration; `scribeConfig` implies a config object, but it's a file, etc). Additionally, lines like `new CacheConfiguration($configName, $configName, true)` are not very expressive.

ArclightHub:
I will split this into two separate PRs

PR #763, opened 2023-11-27
Allow config and temp directories to be configured independently.

ArclightHub:
### Whats changed?

- **Added** the ability to configure the `scribe.php` (or custom php config file) independently of the temp directory.

#### This PR continues https://github.com/knuckleswtf/scribe/pull/760 which has been split into two smaller PRs.

### DIFF
--- a/camel/Camel.php
+++ b/camel/Camel.php
@@ -15,12 +15,12 @@ class Camel
 {
     public static function cacheDir(PathConfig $pathConfiguration): string
     {
-        return $pathConfiguration->getTemporaryDirectoryPath() . "/endpoints.cache";
+        return $pathConfiguration->getTemporaryDirectoryPath('endpoints.cache');
     }
 
     public static function camelDir(PathConfig $pathConfiguration): string
     {
-        return $pathConfiguration->getTemporaryDirectoryPath() . "/endpoints";
+        return $pathConfiguration->getTemporaryDirectoryPath('endpoints');
     }
 
     /**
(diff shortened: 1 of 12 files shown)


shalvah:
Sorry for the delay, merged! 🎉 

This had the key foundations, I just had to rework some of the naming and internal API choices. Sharing a few notes:
- A parameter like `$isHidden` isn't good, because what we want to achieve has nothing to do with hidden/non-hidden, just with automatic inference from the config name. Similarly, I removed some tests focusing on the hidden aspect.
- A method like `getScribeConfigurationPath()` is misleading, because, even though it internally uses the `configName`, one of its purposes is to resolve _output paths_ (for instance, `$pathConfig->getScribeConfigurationPath('openapi.yaml')`). So I split it into separate methods, so this is now `$pathConfig->outputPath('openapi.yaml')`.
- Having repeated parameters and booleans like in `new PathConfig($configName, $configName, true)` is often a suboptimal API, because its unclear to the reader. We can often do better. In this case, I refactored to `new PathConfig($configName)` (since the others depend on the first, by default). For cases of multiple arguments, we can use PHP 8's named arguments to make it clearer:
  ```diff
  - new PathConfig($configName, $this->option('scribe-dir'));
  + new PathConfig(configName: $configName, scribeDir: $this->option('scribe-dir'))
  ```
- It was a good idea to create a base test method (in the `OutputTest`), and have a bunch of test methods call it. However, the exposed API was quite noisy. Rather than having `$this->base(['--a' => 'aa', '--b' =< 'bb'], 'bb')` (unnecessary duplication), this can simply be `$this->base('aa', 'bb')`, and the `base` method takes the key parameters and constructs the needed fields itself. 

I'm sure this was a big effort. Thanks for the initiative!

PR #789, opened 2024-01-18
Enable form validations to support required attributes

Dinushasilva:
This pull request introduces support for validating required form attributes.

**Before**:

theme-elements-before
theme-default-before

**After**:

theme-elements-after
theme-default-after

### DIFF
--- a/resources/css/theme-default.style.css
+++ b/resources/css/theme-default.style.css
@@ -1082,6 +1082,10 @@ button {
     border: none;
 }
 
+.required-attribute {
+    color: #e13939;
+}
+
 * {
     /* Foreground, Background */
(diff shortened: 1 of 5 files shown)


codespearhead:
Can you add a failing test that passes after patch the source code?

shalvah:
Thanks for the suggestion @codespearhead . I'm fine with this as is, since it's a frontend feature. I'm just dubious about accepting the feature, because it's possible people may want to see how the API behaves in cases of missing parameters.

MrCrayon:
> I'm just dubious about accepting the feature, because it's possible people may want to see how the API behaves in cases of missing parameters.

I totally agree, IMHO this form should not validate any input.

PR #792, opened 2024-01-24
Allow multiple content types and schemas (via oneOf) according to OpenAPI 3.0 spec

rhurling:
This PR brings the generator in line with the OpenAPI 3.0 Spec (according to https://swagger.io/docs/specification/describing-responses/) and implements multiple response content types and support for multiple response schemas in a single content type (via oneOf).

Currently missing a Unit test for this behavior (will add this soon).

I couldn't find this specific info anywhere in the documentation so I think there is no change necessary.

As I'm not familiar with Postman I don't know if any change is necessary there.


shalvah:
Completed this, added a test. Although I think it only supports multiple schemas.

PR #804, opened 2024-02-09
Endpoint URL params for OpenAPI

tcox-enrollfirst:
Slight enhancement to the generated OpenAPI specification by simply moving URL parameters down to the endpoint level.  This can improve some http clients' ability to import and render the `openapi.yaml` specification accurately.  Generally, it's probably not a bad idea to define all parameters explicitly at the request level.  Tests have been updated and are passing.

### Example Routes

``` php
/**
 * Search Trucks
 * 
 */
Route::get('trucks', [TruckController::class, 'search'])->name('demo.search');

/**
 * View Truck
 * 
 * @urlParam id string Unique identifier for the truck
 */
Route::get('trucks/{id}', [TruckController::class, 'view'])->name('demo.view');
```

#### openapi.yaml

``` yaml
paths:
  /trucks: ...
  '/trucks/{id}':
    get:
      summary: 'View a Truck'
      operationId: getTrucksId
      description: 'View a specific Truck'
      parameters: []
      responses: ...
      tags: ...
      security: []
    parameters:
      -
        in: path
        name: id
        description: 'The ID of the truck.'
        example: TRK123
        required: true
        schema:
          type: string
```

#### A Client

image

#### openapi.yaml

``` yaml
paths:
  /trucks: ...
  '/trucks/{id}':
    get:
      summary: 'View a Truck'
      operationId: getTrucksId
      description: 'View a specific Truck'
      parameters:
        -
          in: path
          name: id
          description: 'The ID of the truck.'
          example: TRK123
          required: true
          schema:
            type: string
      responses: ...
      tags: ...
      security: []
```

#### Same Client

image

### DIFF
--- a/src/Writing/OpenAPISpecWriter.php
+++ b/src/Writing/OpenAPISpecWriter.php
@@ -98,50 +98,6 @@ class OpenAPISpecWriter
 
             $pathItem = $operations;
 
-            // Placing all URL parameters at the path level, since it's the same path anyway
-            if (count($endpoints[0]->urlParameters)) {
-                $parameters = [];
-                /**
-                 * @var string $name
-                 * @var Parameter $details
-                 */
-                foreach ($endpoints[0]->urlParameters as $name => $details) {
-                    $parameterData = [
-                        'in' => 'path',
-                        'name' => $name,
-                        'description' => $details->description,
-                        'example' => $details->example,
-                        // Currently, OAS requires path parameters to be required
-                        'required' => true,
-                        'schema' => [
-                            'type' => $details->type,
-                        ],
-                    ];
-                    // Workaround for optional parameters
-                    if (empty($details->required)) {
-                        $parameterData['description'] = rtrim('Optional parameter. ' . $parameterData['description']);
-                        $parameterData['examples'] = [
-                            'omitted' => [
-                                'summary' => 'When the value is omitted',
-                                'value' => '',
-                            ],
-                        ];
-
-                        if ($parameterData['example'] !== null) {
-                            $parameterData['examples']['present'] = [
-                                'summary' => 'When the value is present',
-                                'value' => $parameterData['example'],
-                            ];
-                        }
-
-                        // Can't have `example` and `examples`
-                        unset($parameterData['example']);
-                    }
-                    $parameters[] = $parameterData;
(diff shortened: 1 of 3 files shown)


shalvah:
I think this is a bit of a controversial change. Functionality's probably the same, but this will unnecessarily increase the size of specs, no?

tcox-enrollfirst:
It's not unnecessary if the results are more accurate.  Rather be right multiple times than wrong the one time.  Id rather my customers have a few extra lines in a small spec file than a smaller spec file that can be misinterpreted by importers.  In a perfect world, all http clients would load the spec properly but unfortunately that's not the case.

shalvah:
I'm inclined to merge this, but I wouldn't characterize it as "a little more" verbose. If you have a /things/{id} endpoint that can take GET, POST, and DELETE, you'll now have 3x URL parameters rather than 1. Multiply this by the number of variables in the URL (/things/{id}/other-things/{other_id} goes from 2 to 6 variables), and by the number of endpoints (a project that has 50 endpoints of this form will go from 100 variables to 300), and it could really bloat files. But in the first place, the generated specs weren't meant for human consumption, so I think this may be fine.

tcox-enrollfirst:
I hear ya, it's definitely less efficient... Would you be more open to a config-driven solution?

Something like this?
```php
// Generate an OpenAPI spec (v3.0.1) in addition to docs webpage.
// For 'static' docs, the collection will be generated to public/docs/openapi.yaml.
// For 'laravel' docs, it will be generated to storage/app/scribe/openapi.yaml.
// Setting `laravel.add_routes` to true (above) will also add a route for the spec.
'openapi' => [
    'enabled' => true,

    // Generated spec can list URL params at either the "path" or "method" level
    // - "path" lists URL params once for every method that share's the same path
    // - "method" lists URL params under every method within each path grouping
    'url_params' => 'path',

    'overrides' => [
        // 'info.version' => '2.0.0',
    ],
],
```

shalvah:
Honestly, probably not. It makes sense, but I'm wary of adding more config items.

tcox-enrollfirst:
Closing PR as we ultimately decided that this is unnecessary and would cause some bloat in the configs/specs

PR #815, opened 2024-02-26
Fixes #810

### LINKED ISSUE #810: Example Request buggy in Elements Theme
### Scribe version

4.30.1

### PHP version

8.2

### Framework

Laravel

### Framework version

latest

### Scribe config

```ruby
I don't think we need it, but please let us know if we do
```

### What happened?

The example request in the elements theme is truncated if the body is too long and only partially appears after the example response. An example can be seen in the image 

image

A fix would be very helpful in that regard, I think it's just a style problem here. 

Sometimes the two texts also collide with each other

### Docs

MrCrayon:
Adds  `sl-overflow-y-auto` in example request so that content is not truncated but can scroll.

Fixes #810 

Before
scribe_elements_request_before
After
scribe_elements_request_after


shalvah:
Bless you

PR #816, opened 2024-02-27
Fixes : Converting ResponseCalls.php $settings with Extractor::transformOldRouteRulesIntoNewSettings()

VFAndrew:
-using correct array key to access query and body parameters for response calls


shalvah:
Thanks for debugging this! I think I missed a call to this method, so I've added it now, in its proper place. Sorry, gonna close this PR. It should now be fixed in 4.33.0.

PR #819, opened 2024-03-04
feat: write OpenAPI spec in json or yaml format

nathanaelhoun:
From https://swagger.io/specification/, 
> An OpenAPI document that conforms to the OpenAPI Specification is itself a JSON object, which may be represented either **in JSON or YAML format**.

This PR adds a configuration toggle to output the OpenAPI document in the JSON format, which is useful to use it with the package [nextapps-be/laravel-swagger-ui](https://github.com/nextapps-be/laravel-swagger-ui) without the `yaml` PECL extension.

### DIFF
--- a/routes/laravel.php
+++ b/routes/laravel.php
@@ -16,6 +16,6 @@ Route::middleware($middleware)
         })->name('scribe.postman');
 
         Route::get("$prefix.openapi", function () {
-            return response()->file(Storage::disk('local')->path('scribe/openapi.' . config('scribe.openapi.format')));
+            return response()->file(Storage::disk('local')->path('scribe/openapi.' . (config('scribe.openapi.json', false) ? 'json' : 'yaml')));
         })->name('scribe.openapi');
     });
--- a/routes/laravel.php
+++ b/routes/laravel.php
@@ -16,6 +16,6 @@ Route::middleware($middleware)
         })->name('scribe.postman');
 
         Route::get("$prefix.openapi", function () {
-            return response()->file(Storage::disk('local')->path('scribe/openapi.yaml'));
+            return response()->file(Storage::disk('local')->path('scribe/openapi.' . config('scribe.openapi.format')));
         })->name('scribe.openapi');
     });
--- a/src/Writing/Writer.php
+++ b/src/Writing/Writer.php
@@ -95,12 +95,21 @@ class Writer
             c::info('Generating OpenAPI specification');
 
             $spec = $this->generateOpenAPISpec($parsedRoutes);
-            if ($this->isStatic) {
-                $specPath = "{$this->staticTypeOutputPath}/openapi.yaml";
-                file_put_contents($specPath, $spec);
+
+            if (config('scribe.openapi.format') === 'json') {
+                $fileName = "openapi.json";
+                $fileContent = json_encode($spec, JSON_PRETTY_PRINT);
             } else {
-                $outputPath = $this->paths->outputPath('openapi.yaml');
-                Storage::disk('local')->put($outputPath, $spec);
+                $fileName = "openapi.yaml";
+                $fileContent = Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP);
+            }
+
+            if ($this->isStatic) {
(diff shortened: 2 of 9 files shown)


nathanaelhoun:
Thanks for your understanding. I agree that Yaml format is more readable, but allowing to publish as JSON would be a great enhancement for our deploiement setup.

I replaced the configuration for a more simple toggle `scribe.openapi.json` (as it can only be JSON or Yaml according to the OpenAPI spec), and removed it from the default config.

---

Having put more thought to it, I wonder if it wouldn't be simpler for both of us to add a composer script to my side to convert the yaml to json after building. This would avoid merging this new "niche" feature into Scribe. What do you think?

```
php -r "require 'vendor/autoload.php'; echo json_encode(Symfony\Component\Yaml\Yaml::parseFile('resources/swagger/openapi.yaml'), JSON_PRETTY_PRINT);"
```

routes/laravel.php
```diff
@@ -16,6 +16,6 @@
         })->name('scribe.postman');
 
         Route::get("$prefix.openapi", function () {
-            return response()->file(Storage::disk('local')->path('scribe/openapi.yaml'));
+            return response()->file(Storage::disk('local')->path('scribe/openapi.' . (config('scribe.openapi.json', false) ? 'json' : 'yaml')));
```
mikeah2011, line 19: I feel that this will be more unified and elegant `OpenAPISpecWriter::getSpecFileName()`

src/Writing/Writer.php
```diff
@@ -95,12 +95,19 @@ protected function writeOpenAPISpec(array $parsedRoutes): void
             c::info('Generating OpenAPI specification');
 
             $spec = $this->generateOpenAPISpec($parsedRoutes);
+
+            if (config('scribe.openapi.json', false)) {
+                $fileContent = json_encode($spec, JSON_PRETTY_PRINT);
+            } else {
+                $fileContent = Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP);
+            }
```
mikeah2011, line 103: In my personal opinion, writing this part of the logic in generateOpenAPISpec as the return result of a string would be more intuitive, and I am not sure if it is feasible.

src/Writing/Writer.php
```diff
@@ -176,8 +184,8 @@ protected function performFinalTasksForLaravelType(): void
         $contents = preg_replace('#href="\.\./docs/css/(.+?)"#', 'href="{{ asset("' . $this->laravelAssetsPath . '/css/$1") }}"', $contents);
         $contents = preg_replace('#src="\.\./docs/(js|images)/(.+?)"#', 'src="{{ asset("' . $this->laravelAssetsPath . '/$1/$2") }}"', $contents);
         $contents = str_replace('href="../docs/collection.json"', 'href="{{ route("' . $this->paths->outputPath('postman', '.') . '") }}"', $contents);
-        $contents = str_replace('href="../docs/openapi.yaml"', 'href="{{ route("' . $this->paths->outputPath('openapi', '.') . '") }}"', $contents);
-        $contents = str_replace('url="../docs/openapi.yaml"', 'url="{{ route("' . $this->paths->outputPath('openapi', '.') . '") }}"', $contents);
+        $contents = str_replace('href="../docs'.OpenAPISpecWriter::getSpecFileName().'"', 'href="{{ route("' . $this->paths->outputPath('openapi', '.') . '") }}"', $contents);
+        $contents = str_replace('url="../docs'.OpenAPISpecWriter::getSpecFileName().'"', 'url="{{ route("' . $this->paths->outputPath('openapi', '.') . '") }}"', $contents);
```
mikeah2011, line 188: If I'm not mistaken, it should be missing `/`, the correct one should be `../docs/`

mikeah2011 (commented):
Personal suggestion, please refer to.

nathanaelhoun:
After some time and some thoughts about our usage of Scribe, we have decided to go with the [composer script solution](https://github.com/knuckleswtf/scribe/pull/819#issuecomment-1978368284). Less software can be better software.

I'll let someone else re-submit this code if they are still interested in it.

Thanks again for making Scribe, it's a real gamechanger.

shalvah:
Alright then! I'm still fine with adding it if it's hidden away.👍

PR #820, opened 2024-03-08
Fix Sidebar Style Error for 'External_Static' Type and 'Elements' Theme

lz-freedom:
[examples example](https://github.com/stoplightio/elements/blob/main/examples/static-html/zoom-api/index.html)
Based on the code examples from the official repository of 'stoplightio/elements', it appears that our generation template is missing the setting of the 'body' 'height' to '100vh', which causes a styling error in the sidebar. The sidebar height cannot be fixed and changes with the height of the content page on the right. The page is as follows:
image
After the fix:
image


shalvah:
Thanks!

Issue #793, opened 2024-01-25
file_put_contents(public/docs/openapi.yaml): Failed to open stream: No such file or directory


hanspagel:
### Scribe version

4.29.0

### PHP version

 8.2.11

### Framework

Laravel

### Framework version

10.28.0

### Scribe config

```ruby
type => "external_static"
theme => "scalar"
```

### What happened?

It works fine with the default config, but with `external_static` and `scalar` it throws an error when I run `php artisan scribe:generate`:

```
 file_put_contents(public/docs/openapi.yaml): Failed to open stream: No such file or directory

  at vendor/knuckleswtf/scribe/src/Writing/Writer.php:100
     96▕
     97▕             $spec = $this->generateOpenAPISpec($parsedRoutes);
     98▕             if ($this->isStatic) {
     99▕                 $specPath = "{$this->staticTypeOutputPath}/openapi.yaml";
  ➜ 100▕                 file_put_contents($specPath, $spec);
    101▕             } else {
    102▕                 $outputPath = $this->paths->outputPath('openapi.yaml');
    103▕                 Storage::disk('local')->put($outputPath, $spec);
    104▕                 $specPath = Storage::disk('local')->path($outputPath);
```

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
I guess we need to create the public/docs directory first...🤔

mikeah2011:
done

`composer.json`
```
  ... 
  scripts

      "post-install-cmd": [
          "mkdir -p public/docs",
          "@php artisan scribe:generate"
      ],

```

shalvah:
Yeah, would appreciate a PR to fix that in the package.

mikeah2011:
done https://github.com/knuckleswtf/scribe/pull/829

PR #829, opened 2024-03-18
make directory recursive

mikeah2011:
Fixes https://github.com/knuckleswtf/scribe/issues/793 

Referring to other creation methods, it was found that a method should be encapsulated for calling.

Additionally, remove the Str class that is not referenced in the Write class.


shalvah:
Thanks!

PR #830, opened 2024-03-20
added cast number value to float if value is string

ptelevich:
Screenshot 2024-03-20 at 16 33 48
Screenshot 2024-03-20 at 16 35 53


ptelevich:
Eventually the fix cast all strings with type number (integer, float/double) to float, whether this is int or float. I suppose it's better than get string on backend side

shalvah:
Great, thank you!

PR #739, opened 2023-10-02
[update] nullable Parameter For OpenApi3

liverocaisson:
If not required field 
allow nullable parameter 
as OpenAPI suggest

PR #834, opened 2024-03-28
feature/nullable

MertcanDinler:
Support for nullable values for OpenApiSpecs.
If the parameter is not required, if it is specified to be nullable with the BodyParam attribute, or if there is a nullable rule in the validation, this field will be added to the OpenApiSpecs as nullable.

### DIFF
--- a/camel/Extraction/Parameter.php
+++ b/camel/Extraction/Parameter.php
@@ -14,6 +14,7 @@ class Parameter extends BaseDTO
     public string $type = 'string';
     public array $enumValues = [];
     public bool $exampleWasSpecified = false;
+    public bool $nullable = false;
 
     public function __construct(array $parameters = [])
     {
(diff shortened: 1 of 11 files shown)


liverocaisson:
@MertcanDinler cc @shalvah 
I tried also to implement a similar nullable fonctionnality when not required -> nullable : true
but tests don't pass. If someone can help. 
https://github.com/knuckleswtf/scribe/pull/739

jfoliveira:
> @MertcanDinler cc @shalvah I tried also to implement a similar nullable fonctionnality when not required -> nullable : true but tests don't pass. If someone can help. #739

@liverocaisson added more suggestions to #739
Please have a look when you have a chance.

shalvah:
Will take a look at these sometime this week or next.

shalvah:
Hey, sorry I didn't get back to this! I think I'll accept it. Note that, as you've PRed, this feature only works with the attribute and the validation rule, and only affects the OpenAPI output. This means:
1. It doesn't work with the annotations (`@bodyParam`, ...)
2. It doesn't change the HTML page output.

I think I'd like to change the second, but the first will probably stay as is for a while.  So I'll accept this, but label it "experimental".

PR #840, opened 2024-04-30
Array with Form request doesn't generate both example and fields at the same time

wcostaprijo:
When using it in the form request:

Captura de tela de 2024-04-30 14-04-08

It generates the examples, but it doesn't generate the fields in the body parameters.

Captura de tela de 2024-04-30 14-04-46

If I remove the parent index:

Captura de tela de 2024-04-30 14-06-55

It generates the fields in the body parameters, but it doesn't generate the examples.

Captura de tela de 2024-04-30 14-07-53

After the correction, passing the parent index, it generates the examples and also generates the fields in the body parameters.

Captura de tela de 2024-04-30 14-04-08
Captura de tela de 2024-04-30 14-09-40


shalvah:
Nice, thanks.

Issue #836, opened 2024-04-08
Annotation example for boolean values doesn't work correctly


BnitoBzh:
### Scribe version

4.35.0

### PHP version

8.3.3

### Framework

Laravel

### Framework version

10.43.0

### Scribe config

```ruby
type => "laravel"
theme => "elements"
static.output_path => "public/api/docs"
laravel.docs_url => "/api/docs"
auth.enabled => true
auth.default => true
auth.name => "apiKey"
```

### What happened?

Attribute annontations do not render boolean examples correctly.

`#[BodyParam('newsletter', 'boolean', description: 'Opt-in status for newsletter', required: true, example: true)]`
`#[BodyParam('newsletter', 'boolean', description: 'Opt-in status for newsletter', required: true, example: false)]`

With `true` or `false` value as example, there is no example text in the documentation and for `false`value the curl example is correct, but with the `true` value, the curl example is set to `null`

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

mikeah2011:
`#[QueryParam('is_api_user', 'boolean', '聯盟站點是否異業串接', false, 'true', ['true-是', 'false-否'])]
`

BnitoBzh:
Sorry @mikeah2011 i am speaking about BodyParam not QueryParam.
If the example for a boolean BodyParam is set as string, like your comment `'true'`, the API json body sample use the type string instead of the type boolean.
In my example, the sample will be : `{ "newsletter": "true" }` instead of `{ "newsletter": true }`

mikeah2011:
Yes, the string true cannot solve the problem perfectly. I have made a correction, please refer to it.

PR #841, opened 2024-05-01
Hotfix/fix type comparison

mikeah2011:
https://github.com/knuckleswtf/scribe/issues/836

After investigation, it was found that weak type comparison was used due to string comparison, and using `===` can solve the problem


shalvah:
Fucking PHP. Thanks.

PR #843, opened 2024-05-01
Hotfix/fix type comparison

mikeah2011:
https://github.com/knuckleswtf/scribe/issues/836

Sorry, I should modify the blade file together to ensure that it is displayed correctly.

If not modified in this way, `example` will display 0/1, but I did not make any hasty changes for `example-requests`.

I am thinking about how to write test cases that are more convincing. Let me take a screenshot and see the diagram below:

image


mikeah2011:
Temporarily closed

PR #847, opened 2024-05-06
feat: add afterResponseCall hook

max13fr:
Add the hook **afterResponseCall** allowing to modify the result after the API call (for example to reduce large examples with hundred of lines to 5 maximum, to remove a sensible data, ...).

The doc update is available here : https://github.com/knuckleswtf/scribe-docs/pull/29

### DIFF
--- a/src/Extracting/Strategies/Responses/ResponseCalls.php
+++ b/src/Extracting/Strategies/Responses/ResponseCalls.php
@@ -6,7 +6,6 @@ use Dingo\Api\Dispatcher;
 use Dingo\Api\Routing\Route as DingoRoute;
 use Exception;
 use Illuminate\Contracts\Http\Kernel;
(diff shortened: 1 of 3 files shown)


src/Extracting/Strategies/Responses/ResponseCalls.php
```diff
@@ -170,6 +174,13 @@ protected function runPreRequestHook(Request $request, ExtractedEndpointData $en
         }
     }
 
+    protected function runPostRequestHook(Request $request, ExtractedEndpointData $endpointData, JsonResponse $jsonResponse): void
```
shalvah, line 177: Is the response always of type JSONResponse? I don't think we should add that constraint.

max13fr: Oh yes, my bad, I fixed it

PR #855, opened 2024-05-19
Introduce Adding examples to the OAS3 Media Type Object (Request / Response)

Dinushasilva:
This PR introduces to create an entire request or response examples by leveraging existing OAS3 capabilities by allowing multiple examples with arbitrary names.

Having multiple examples will increase user experience as it allows consumers to see different ways that an API can be consumed / requested.

This is by default supported in most documentation tooling to help users pick which example they'd like to see. We could also expand examples into Objects and Parameters at a later stage.

### Before:
Screenshot 2024-05-18 163231

### After:

Scalar:

image

image

Rapidoc

image

image

Stoplight

image

image

Swagger

image

image

I can update the docs, if we can agree on this implementation.

Thank you!

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -248,6 +248,9 @@ INTRO
             Strategies\ResponseFields\GetFromResponseFieldAttribute::class,
             Strategies\ResponseFields\GetFromResponseFieldTag::class,
         ],
+        'examples' => [
+            Strategies\Examples\UseExampleTag::class,
+        ],
     ],
 
     // For response calls, API resource responses and transformer responses,
--- a/tests/Fixtures/TestController.php
+++ b/tests/Fixtures/TestController.php
@@ -527,7 +527,6 @@ class TestController extends Controller
 
         // Do stuff
         if ($validator->fails()) {
-
         }
     }
 
@@ -585,7 +584,7 @@ class TestController extends Controller
     {
         return null;
     }
-    
+
     public function withInjectedModelFullParamName(TestPost $testPost)
     {
         return null;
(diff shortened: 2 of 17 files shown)


config/scribe.php
```diff
@@ -248,6 +248,9 @@
             Strategies\ResponseFields\GetFromResponseFieldAttribute::class,
             Strategies\ResponseFields\GetFromResponseFieldTag::class,
         ],
+        'examples' => [
+            Strategies\Examples\UseExampleTag::class,
```
shalvah, line 252: Nice. I feel like attributes are a more convenient way to specify multiple examples, so let's add support for that too.

shalvah (commented):
Still need to look at it in more detail, but this looks promising

config/scribe.php
Dinushasilva: @shalvah let me know how you would like me to update the docs, should it be a new section under Eg: [Documenting your API](https://scribe.knuckles.wtf/laravel/documenting/#)

shalvah: Yeah, that makes sense.

shalvah: I'd say you can probably add it under the existing docs for requests and responses.

shalvah:
Been busy, will take a look at these sometime this week or next.

tests/Fixtures/TestController.php
```diff
@@ -608,6 +607,116 @@ public function withInjectedEnumAndModel(Category $category, TestUser $user)
         return null;
     }
      */
+
+    /**
+     * @example
```
shalvah, line 612: I think I'd rather change this to two separate tags, `@exampleRequest` and `@exampleResponse`. My reasoning is that `@example` could be a generic DocBlock tag, and I'd prefer to avoid any conflicts.

tests/Fixtures/openapi.yaml
```diff
@@ -176,6 +176,54 @@ paths:
...
+                    content:
+                        text/plain:
+                            schema:
+                                type: string
+                                example: ''
+            tags:
+                - 'Group A'
+            security: []
+    '/api/withExampleTagTypeRequest':
+        get:
```
shalvah, line 204: Hmmm, doesn't seem like the contents match what should be generated here. I don't see any example requests.

shalvah (commented):
Thanks for your contribution! This is a nice idea, and thanks for the screenshots. But my concern: How does this fit with the other response methods, especially when you throw in the `scenario` feature? For instance, the response examples in your test cases are essentially the same as writing `@response 200 <...>`.

When I think about it, the existing `@response-` family of tags and annotations are already for specifying example responses, it's pretty much identical to these.

My current view: 
1. Rather than a generic `@example` tag, let's use the `@response` tag and add a `@request` tag, to align with the `@response` tag.
2. Let's rather ensure the multiple example requests/responses are correctly written in the OpenAPI spec (I think that's what https://github.com/knuckleswtf/scribe/pull/792 is about).

shalvah:
FYI, I've merged https://github.com/knuckleswtf/scribe/pull/792, which I think implements some of this functionality.

Issue #764, opened 2023-12-01
When using FormRequest with 'file' validation, request data is empty even data is provided in docs example body


lHondol:
### Scribe version

4.23

### PHP version

8.2.4

### Framework

Laravel

### Framework version

10

### Scribe config

```ruby
theme => "elements"
title => "Notex"                  
base_url => "http://localhost:8000"           
try_it_out.base_url => "http://localhost:8000"
auth.enabled => true
auth.default => true
auth.name => "Authorization"
```

### What happened?

In my FormRequest, i have these rules:

    public function rules(): array
    {
        return [
            'information' => ['required'],
            'file' => ['required', 'file'],
        ];
    }

In the generated documentation, within the "try_out" body, I can now choose the file I need to upload. However, when I hit submit, the response states, 'The information field is required,' even though I have already provided the value.

also the request payload:
```
------WebKitFormBoundaryPuRQG2gkhYZdgo8K
Content-Disposition: form-data; name="information"


tempore

------WebKitFormBoundaryPuRQG2gkhYZdgo8K
Content-Disposition: form-data; name="file"; filename="myfile.pdf"
Content-Type: application/pdf


------WebKitFormBoundaryPuRQG2gkhYZdgo8K--
```

request headers :
Content-Type: 'multipart/form-data'
Accept: 'application/json'[](url)

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

mcpicoli:
Having similar trouble here using try-it-out. Route uses HTTP PUT method and has a "file" parameter named "photo". ```php artisan scribe:generate``` generates "correct" documentation, I can pick a file and fill other parameters. If no file is selected, everything works fine, only when providing it, it is missing as described by the OP.

#257 should have fixed this problem, but the HTTP method is never changed to POST, the hidden parameter "_method" is never added and the Content-Type header is never changed to multipart/form-data. The file is never seen by the server (as expected since PUT does not support multipart/form-data in PHP) and because of that the validation logic fails, stating the file is invalid since it is encoded as an empty JSON body.

```{"photo":{}}```

The request header, as per the browser developer tools is not changed from "Content-Type: application/json"

[...]

PR #864, opened 2024-07-09
fix: fixes elements theme mutlipart form file upload

wlhrtr:
When uploading a file the Content-Type header must not be set otherwise the browser cannot set the right boundaries for the multipart formdata payload.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects
Bildschirmfoto 2024-06-06 um 12 07 56

TryItOut editor in `elements` theme is currently setting the Content-Type header and therefore file uploads do not work.

Should resolve https://github.com/knuckleswtf/scribe/issues/764

### DIFF
--- a/resources/views/themes/elements/index.blade.php
+++ b/resources/views/themes/elements/index.blade.php
@@ -111,6 +111,11 @@
                     });
                 }
 
+                // content type has to be unset otherwise file upload won't work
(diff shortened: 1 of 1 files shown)


shalvah:
Thanks!

Issue #867, opened 2024-07-11
Response fields overflows to dark box on default theme


dfsmania:
### Scribe version

4.37.1

### PHP version

8.3.9

### Framework

Laravel

### Framework version

10.48.16

### Scribe config

```ruby
title => "My API"
static.output_path => "public/docs/api"
auth.enabled => true
auth.default => true
auth.name => "token"
auth.placeholder => "{YOUR_ACCESS_TOKEN}"
logo => "../../images/app-logo.png"
strategies.urlParameters => added Knuckles\Scribe\Extracting\Strategies\UrlParameters\GetFromLumenAPI
strategies.headers => added Knuckles\Scribe\Extracting\Strategies\Headers\GetFromRouteRules: removed [
    'override',
    [
        'Content-Type' => 'application/json',
        'Accept' => 'application/json',
    ],
]
strategies.responses => added Knuckles\Scribe\Extracting\Strategies\Responses\ResponseCalls: removed [
    'Knuckles\\Scribe\\Extracting\\Strategies\\Responses\\ResponseCalls',
    [
        'only' => [
            'GET *',
        ],
        'config' => [
            'app.debug' => false,
        ],
    ],
]
```

### What happened?

I ran `php artisan scribe:generate` and the output documentation using the **default theme** presents an issue with the **response fields**, the text of the descriptions overflows to the `dark-box` zone. Check next image:

![image](https://github.com/knuckleswtf/scribe/assets/63609705/e96e2385-24fc-4597-85ae-4753db2a3cc6)

Using the inspector tool of the browser, I can see that other fields, like the **body parameters** are wrapped inside a `div` holding the next style:

```css
margin-right: 50%
```

While the `div` that wraps those response fields are missing that extra style.

### Docs

- [X] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
Thanks for investigating!

PR #868, opened 2024-07-11
Fix issue 857 (Response fields overflows to the dark box zone)

dfsmania:
Fix issue #867 and remove some duplicated code on `resources/css/theme-default.style.css`


shalvah:
I have no idea if this works, but I'll take your word for it, lol.

dfsmania:
@shalvah it worked on my local environment. However, I'm using the `elements` theme now. So I tested that change for about two weeks only.

PR #870, opened 2024-07-16
fixed for making instances with dep. injected

Tankonyako:


shalvah:
That's a significant change, bro. Gotta explain your reasoning. And check history of other discussions here.

PR #874, opened 2024-07-26
Show enum case name in output

Lea-Reif:
In cases where BackedEnums have int type, the output is just a list of numbers, ex:
image
With this changes, enums with int type will show the base name of the case:
image

### DIFF
--- a/resources/views/themes/elements/components/field-details.blade.php
+++ b/resources/views/themes/elements/components/field-details.blade.php
@@ -38,7 +38,7 @@
         @endif
         @if(!empty($enumValues))
             Must be one of:
-            <ul style="list-style-position: inside; list-style-type: square;">{!! implode(" ", array_map(fn($val) => "<li><code>$val</code></li>", $enumValues)) !!}</ul>
+            <ul style="list-style-position: inside; list-style-type: square;">{!! implode(" ", array_map(fn($val, $key) => "<li><code>$val ($key)</code></li>", $enumValues, array_keys($enumValues))) !!}</ul>
         @endif
         @if($isArrayBody)
             <div class="sl-flex sl-items-baseline sl-text-base">
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -2,6 +2,7 @@
 
 namespace Knuckles\Scribe\Extracting;
 
+use BackedEnum;
 use Illuminate\Contracts\Validation\Rule;
 use Illuminate\Contracts\Validation\ValidationRule;
 use Illuminate\Support\Arr;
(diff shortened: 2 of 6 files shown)


shalvah:
👀

I didn't get the chance to look at this, but it seems like a good idea. Still interested?

Lea-Reif:
If you can't reopen it, I would gladly redo it

src/Extracting/Strategies/GetFromInlineValidatorBase.php
```diff
@@ -90,7 +90,7 @@ enum_exists($enum) && method_exists($enum, 'tryFrom')
                         // $case->value only exists on BackedEnums, not UnitEnums
                         // method_exists($enum, 'tryFrom') implies $enum instanceof BackedEnum
                         // @phpstan-ignore-next-line
-                        $rulesList[] = 'in:' . implode(',', array_map(fn ($case) => $case->value, $enum::cases()));
+                        $rulesList[] = 'in:' . implode(',', array_map(fn ($case) => "$case->value ($case->name)", $enum::cases()));
```
shalvah, line 93: I'm deleting this line because it's not needed. The rulesList is the actual list of rules that Laravel supports, and changing this from `in:1,2,3` to `in:1 (MALE),2 (FEMALE),3 (UNDEFINED)` will probably break things.

src/Extracting/ParsesValidationRules.php
```diff
@@ -208,9 +209,13 @@ protected function parseRule($rule, array &$parameterData, bool $independentOnly
...
-                // @phpstan-ignore-next-line
-                $cases = array_map(fn ($case) => $case->value, $type::cases());
-                $parameterData['type'] = gettype($cases[0]);
+                $cases = [];
+                foreach ($type::cases() as $case) {
+                    /** @var BackedEnum $case */
+                    $cases[$case->name] = $case->value;
+                }
+                
+                $parameterData['type'] = gettype(reset($cases));
```
shalvah, line 218: Hmm, what does this do? The `$cases` array is not used for anything else beyond getting the type, and it does the same thing as the previous version, which works for Backed Enums.

resources/views/themes/elements/components/field-details.blade.php
```diff
@@ -38,7 +38,7 @@ class="svg-inline--fa fa-chevron-right fa-fw fa-sm sl-icon" role="img"
         @endif
         @if(!empty($enumValues))
             Must be one of:
-            <ul style="list-style-position: inside; list-style-type: square;">{!! implode(" ", array_map(fn($val) => "<li><code>$val</code></li>", $enumValues)) !!}</ul>
+            <ul style="list-style-position: inside; list-style-type: square;">{!! implode(" ", array_map(fn($val, $key) => "<li><code>$val ($key)</code></li>", $enumValues, array_keys($enumValues))) !!}</ul>
```
shalvah, line 41: How does this render for non-backed enums? This looks like it would result in "0 (0), 1 (1)".

shalvah:
I can't reopen it, as the repository has been deleted. 😞 Please redo if you can (but see my comments).

PR #880, opened 2024-08-07
Fix #879 (Comments are ignored for validator parameters with non string/array (e.g. conditional) rule lists)

### LINKED ISSUE #879: Comments are ignored for validator parameters with non `string`/`array` (e.g. conditional) rule lists
### Scribe version

4.37.1

### PHP version

8.3.7

### Framework

Laravel

### Framework version

11.16

### Scribe config

```ruby
N/A
```

### What happened?

When validator rules contain conditional values (as with `another_val` in the example below), the `GetFromInlineValidatorBase` strategy does not parse the rules, which is expected as only a basic string/array of validator rules are supported. 

```php
$accepts_another_val = true;

$request->validate([
  // A comment describing some_val. Example: null
  'some_val' => ['required', 'max:255', 'string'],
  // A comment describing another_val, only required when certain condition is met. Example: null
  'another_val' => $accepts_another_val ? ['required', 'max:255',  'string'] : ['prohibited'],
]);
```

However, the preceding comment is also ignored and is not included in the generated documentation. This is due to `GetFromInlineValidatorBase` bailing when the validator rule is not a `PhpParser\Node\Scalar\String_` or `PhpParser\Node\Expr\Array_` (in the example above the value is a `PhpParser\Node\Expr\Ternary`), _before_ applying any data from comments:

https://github.com/knuckleswtf/scribe/blob/5eb0f65973db9df5ba455f8bbcc8a1cd6573564b/src/Extracting/Strategies/GetFromInlineValidatorBase.php#L97-L100

I think it should be safe to remove the `continue` here? In which case only supported values (basic `string`/`array`) will be parsed by the strategy, but it would be possible for us to add context via the preceding comment. 

### Docs

thomasfw:
Fixes #879

### DIFF
--- a/src/Extracting/Strategies/GetFromInlineValidatorBase.php
+++ b/src/Extracting/Strategies/GetFromInlineValidatorBase.php
@@ -96,7 +96,6 @@ class GetFromInlineValidatorBase extends Strategy
                 $rules[$paramName] = join('|', $rulesList);
             } else {
                 $rules[$paramName] = [];
-                continue;
             }
 
             $dataFromComment = [];


shalvah:
Makes sense, thanks!

PR #884, opened 2024-08-23
Allow custom output path for `static` and `external_static` instead of only `static`

PatryQHyper:
When you set type to `external_static` and set custom static output_path, it still points to default path in generated HTML.

Generated part of scalar HTML (before PR, output_path is different from `public/docs`):
```html
<script
    id="api-reference"
    data-url="../docs/openapi.yaml">
</script>
```


shalvah:
Good catch!

PR #887, opened 2024-09-05
Fix display of boolean examples on the elements theme

dfsmania:
This **PR** fix the rendering of boolean examples with the **elements theme**. Currently, there's an issue and boolean example cases like `false` and `true` aren't rendering in the right way. At next you can see an image showing this issue:

### Previous to this PR
Screenshot 2024-09-05 at 11-27-01 The Giga Provisioning API

In this image you can see that a `false` example value is not rendered and `true` is rendered like `1`.

### After this PR
Screenshot 2024-09-05 at 11-27-52 The Giga Provisioning API

In this image you can see that boolean example cases are now rendering nicely.

### DIFF
--- a/resources/views/themes/elements/components/field-details.blade.php
+++ b/resources/views/themes/elements/components/field-details.blade.php
@@ -49,12 +49,12 @@
                 @endif
             </div>
         @endif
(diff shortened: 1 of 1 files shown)


shalvah:
Thanks!

PR #888, opened 2024-09-06
Postman collection export: encoding value to prevent validation errors

eschricker:
The generated postman collection from my project was not valid. It contained several `value` entries for query parameters that were pure integer values. This is not compatible with the postman format, which requires strings.

Therefore, I added `urlencode` to the `value` in `PostmanCollectionWriter` to avoid validation errors.

### DIFF
--- a/src/Writing/PostmanCollectionWriter.php
+++ b/src/Writing/PostmanCollectionWriter.php
@@ -293,7 +293,7 @@ class PostmanCollectionWriter
                     // See https://www.php.net/manual/en/function.parse-str.php
                     $query[] = [
                         'key' => "{$name}[$index]",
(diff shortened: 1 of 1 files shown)


shalvah:
Hm, I'm reluctant to add this because URL encoding will potentially have other effects on regular strings. I think there's been a lot of back and forth on encoding for Postman collections in this project. People complain when we don't encode; people complain when we do.

eschricker:
Would it be better to check if the value is a string to take it as it is and only encode if it is not a string?

shalvah:
Yeah, let's do that.

eschricker:
I have changed `urlencode` with `strval` due to complains of phpstan. `urlencode` expects a string as input.

PR #893, opened 2024-09-19
Improve Laravel dd() Output Rendering in Scribe Documentation

sabist:
This PR addresses an issue where the output of Laravel's dd() function is rendered as plain text in the generated documentation by Scribe. The changes introduce a selective rendering mechanism that safely treats dd() output as HTML while keeping all other API responses as plain text, ensuring proper security.

- Laravel dd() detection: The script now checks for the presence of sf-dump in the response to identify dd() output.

- Secure rendering: Only the detected dd() output is rendered using innerHTML to display the HTML dump correctly. Other responses continue to use textContent to avoid XSS vulnerabilities.

This update improves the developer experience by ensuring that Laravel debug outputs are correctly formatted in the API documentation, while maintaining security.

### DIFF
--- a/resources/css/theme-default.style.css
+++ b/resources/css/theme-default.style.css
@@ -894,6 +894,11 @@ html {
     text-shadow: 0 1px 2px rgba(0, 0, 0, .4)
 }
 
+.content blockquote pre.sf-dump,
+.content pre pre.sf-dump {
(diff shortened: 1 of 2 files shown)


shalvah:
Good one, thanks!

Issue #620, opened 2023-02-20
Support `withCount` in `ResponseFromApiResource`


gwlortscher:
When using the `ResponseFromApiResource` attribute, it's possible to account for usage of [whenLoaded](https://laravel.com/docs/master/eloquent-resources#conditional-relationships) in Eloquent API Resources using the `with:` argument to eager load relationships. `whenLoaded` is helpful in preventing N+1 query problems in API Resources. 

[whenCounted](https://laravel.com/docs/master/eloquent-resources#conditional-relationship-counts) / [withCount](https://laravel.com/docs/master/eloquent-relationships#counting-related-models) is similarly helpful in preventing N+1 query problems, but doesn't appear to be supported by Scribe. This means counts won't appear in API docs generated by Scribe. 

A good solution could be to support `withCount` in `ResponseFromApiResource`?

shalvah:
I don't work much with these features, so I wouldn't know. PRs welcome!

thewhaleiam:
> When using the `ResponseFromApiResource` attribute, it's possible to account for usage of [whenLoaded](https://laravel.com/docs/master/eloquent-resources#conditional-relationships) in Eloquent API Resources using the `with:` argument to eager load relationships. `whenLoaded` is helpful in preventing N+1 query problems in API Resources.
> 
> [whenCounted](https://laravel.com/docs/master/eloquent-resources#conditional-relationship-counts) / [withCount](https://laravel.com/docs/master/eloquent-relationships#counting-related-models) is similarly helpful in preventing N+1 query problems, but doesn't appear to be supported by Scribe. This means counts won't appear in API docs generated by Scribe.
> 
> A good solution could be to support `withCount` in `ResponseFromApiResource`?

Was there any progress made on this?

PR #920, opened 2024-11-18
Add withCount support

SergeAntonets:
- See: https://github.com/knuckleswtf/scribe/issues/620
- Laravel: https://laravel.com/docs/6.x/eloquent-relationships#counting-related-models
- The "whenCounted" method was introduced in Laravel v9. See: https://github.com/laravel/framework/pull/43101

This PR adds **withCount** to the **ResponseFromApiResource** and will be supported by the **factoryCreate** model source


src/Attributes/ResponseFromApiResource.php
```diff
@@ -23,6 +23,7 @@ public function __construct(
         public ?int $paginate = null,
         public ?int $simplePaginate = null,
         public array $additional = [],
+        public array $withCount = [],
```
SergeAntonets, line 26: I am adding it at the end so it does not break the backward compatibility

SergeAntonets:
Hi @shalvah, I am finalizing the PR and want to add some test coverage, but it actually requires adding a database to test the "withCount" feature. Additionally, I don't see any tests related to the "with" feature, likely because it also requires a database, which might be why it's missing. 

Any thouths/suggestions around this? thanks

shalvah:
There are some tests that do use a database (SQLite in memory). We even have migrations. Example: https://github.com/knuckleswtf/scribe/blob/4d4a44e48afd4f1f236c9e75cf3cbe8b6b515eb3/tests/Unit/ValidationRuleParsingTest.php#L45-L47

You can follow the example.

SergeAntonets:
@shalvah, thanks. I missed that. The PR is ready; please take a look whenever you have time.

shalvah:
Is the `withCount()` method not supported on older Laravel versions? If so, you can skip the test using something like https://github.com/knuckleswtf/scribe/blob/46e8399eddc1e1b6c2ba31a1f78d6f301483cf74/tests/Strategies/GetFromInlineValidatorTest.php#L236. I've forgotten how, but Laravel provides a method to get the framework version.

SergeAntonets:
@shalvah fixed! The "whenCounted" method was introduced in Laravel v9. See: https://github.com/laravel/framework/pull/43101

PS: added a few more checks

PR #922, opened 2024-11-27
Php 84 upgrade

markwalet:
The dev dependencies were a bit outdated. I've updated them where I could without breaking tests, but especially in the lower versions it was unclear to me which versions were still supported.

Besides the dependency upgrades, I've resolved the deprecations warnings for PHP 8.4.


shalvah:
Thanks!

I'll be fine with dropping support for PHP 8.0 and dingo/api; I think we've done a good job of supporting them for a while now.

markwalet:
Any update on this? What do you want me to do?

markwalet:
I've removed the dependency changes completely in favor of just doing the code changes in a separate merge request to improve reviewability.

https://github.com/knuckleswtf/scribe/pull/929

shalvah:
I've been in the hospital/on recovery. I'll look at this soon. My plan is to create a branch for the next major version and do the ground work there, then point your PRs to it.

shalvah:
Hmm, I think I may close this PR. I've dropped support for PHP 8, Dingo and Lumen on the `vNext` branch, and pointed this to it, but some tests still fail. Current dependencies seem to work fine on PHP 8.1 to 8.4.

shalvah:
Okay, ,they seem to work on all versions but 8.1 (code coverage is broken). Could you check that?

It's not really a priority, anyway, since it's only dev deps.

markwalet:
Unfortunately I can't get the dependencies to work. It's a bit of a mess with a separate lowest.composer.json and I can't wrap my head around it.

Even with the tests being successful, it seems like the coverage statistics are failing. On my own packages I normally only collect coverage data for the latest stable php/laravel version. Maybe that could help? It all just becomes a bit more complicated when you try to run coverage on pestphp V1 **and** V2.

shalvah:
Yep, sorry. Ultimately, there wasn't a real need for this, but thanks for the effort.

PR #932, opened 2025-01-06
Add support for defining Groups and Subgroups as enums

lioneaglesolutions:
## Problem
When defining groups and subgroups, it's nice to have a clear idea of what groups are available to use. PHP Backed Enums are a perfect way of achieving this. Otherwise, developers could mis-type group names easily such as;
```php
#[Group('Users')];
#[Group('User')];  // typo
```

## Solution
By allowing the usage of Backed Enums, it is very clean to be able to define your API Endpoint Groups as an enum;
```php
enum EndpointGroup: string
{
    case Users = 'Users';
    case Invoices = 'Invoices';
}
```
Then in your codebase you can simply do;
```php
#[Group(EndpointGroup::Invoices)]
class InvoiceController
{
   // ...
}
```

### Alternative
An alternative could be to just add `->value` to each enum which would not require any codebase changes however this is not visually pleasing and not as nice of a DX as being able to pass an enum directly.
```php
#[Group(EndpointGroup::Invoices->value)]
class InvoiceController
{
   // ...
}
```

### DIFF
--- a/src/Attributes/Subgroup.php
+++ b/src/Attributes/Subgroup.php
@@ -8,15 +8,30 @@ use Attribute;
 class Subgroup
 {
     public function __construct(
-        public string $name,
+        public mixed $name,
         public ?string $description = '',
     ){
+    }
+
+    protected function getName(): string
+    {
+        if (is_string($this->name)) {
+            return $this->name;
+        }
+
+        if (interface_exists('BackedEnum') && is_a($this->name, 'BackedEnum')) {
+            return $this->name->value;
+        }
+
+        throw new \InvalidArgumentException(
+            'The name property of a group must be either a PHP Backed Enum or a string'
+        );
     }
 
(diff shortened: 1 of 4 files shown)


shalvah:
Nice convenience feature!

src/Attributes/Subgroup.php
```diff
@@ -8,15 +8,30 @@
 class Subgroup
 {
     public function __construct(
-        public string $name,
+        public mixed $name,
         public ?string $description = '',
     ){
     }
 
+    protected function getName(): string
```
shalvah, line 16: Ugh, I really hate these property accessor methods.

Considering making the next major PHP 8.4 only so I can do this with [property hooks](https://www.php.net/manual/en/language.oop5.property-hooks.php).

(I won't do that.)

PR #944, opened 2025-02-08
fix: allow blade syntax in baseUrl config

### LINKED ISSUE #943: Dynamic String not escape properly when using elements theme
### Scribe version

 4.39.0

### PHP version

8.3.16

### Framework

Laravel

### Framework version

10.48.25

### Scribe config

```ruby
base_url => "{{ config(\"app.tenant_url\") }}"
type => "laravel"
theme => "elements"
try_it_out.use_csrf => true
auth.enabled => true
auth.default => true
```

### What happened?

I change the base url to use dynamic string like `"{{ config(\"app.tenant_url\") }}"` according to the documentation and change the theme to `elements` and the dynamic is not escape properly 

Image

### Docs

- [x] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

chengkangzai:
Fixes #943

This PR allows the use of Blade syntax in the `baseUrl` configuration by rendering the value through the Blade facade before displaying it.

### Changes
- Added Blade facade import
- Wrapped `baseUrl` values in `Blade::render()` to process any Blade syntax

### Notes
- Tested with Laravel only, compatibility with Dingo API has not been verified
- This enables dynamic baseUrl values using Blade syntax like `{{ config('app.url') }}` or `{{ config('app.tenant_url') }}`

### References
- Issue #943: Support blade syntax in config baseUrl

Let me know if any changes are needed!

### DIFF
--- a/resources/views/themes/elements/endpoint.blade.php
+++ b/resources/views/themes/elements/endpoint.blade.php
@@ -1,6 +1,7 @@
 @php
     use Knuckles\Scribe\Tools\Utils as u;
     /** @var  Knuckles\Camel\Output\OutputEndpointData $endpoint */
+    use Illuminate\Support\Facades\Blade as Blade;
 @endphp
 
 <div class="sl-stack sl-stack--vertical sl-stack--8 HttpOperation sl-flex sl-flex-col sl-items-stretch sl-w-full">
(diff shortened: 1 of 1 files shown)


shalvah:
Can you change this to do the render call before pasing it to the view? That way, it works for all themese, not just `elements`.

PR #951, opened 2025-02-19
fix: Undefined array key "properties"

SRWieZ:
Hey! An happy Scribe user here.

I just tried updating to version 5 and encountered an exception while generating the documents.

I'm not sure about this PR. I haven't explored it enough to understand how the internals work but this PR did resolve my issue.

Please treat this PR as a bug report and don't hesitate to close it down.

Capture d’écran 2025-02-19 à 12 09 56


shalvah:
Thank you, I'll add a failing test for this.

SRWieZ:
Perfect. Thanks again for Scribe.

Issue #949, opened 2025-02-18
How to configure Scalar


TS1997:
### Scribe version

4.39.0

### Your question

How do I configure the parameters that exist [here](https://github.com/scalar/scalar/blob/main/documentation/configuration.md#hidetestrequestbutton-boolean) for the Scalar implementation? I looked at the view which iterates through `external->html_attributes` and tried entering the following properties one by one but I couldn't get any of them to work. 
```
html_attributes => [
  'hideTestRequestButton' => true,
  'hideTestRequestButton' => "true",
  'hide-test-request-button' => true,
  'hide-test-request-button' => "true",
  'data-hide-test-request-button' => true,
  'data-hide-test-request-button' => "true",
]
```

What's the correct way of entering these configuration values?

### Docs

- [x] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

TS1997:
@shalvah I've done some additional testing of this now and can't get it to work. If I set the configuration like this I get an array to string conversion error in the blade template:
```php
'external' => [
        'html_attributes' => [
            'data-configuration' => [
                'hideTestRequestButton' => true,
            ]
        ]
    ],
```

and if I try to work around that by encapsulating it in a json_encode like this I get an error in the browser console stating that there's a syntax error 
```php
'external' => [
        'html_attributes' => [
            'data-configuration' => json_encode([
                'hideTestRequestButton' => true,
            ])
        ]
    ],
```

![Image](https://github.com/user-attachments/assets/c7adbf8a-9304-4e11-a985-e1a72c945e2d)

As I see it there's two ways to go about fixing this:

1. Adding a check in the blade template that does a JSON.stringify on html_attribute values that are arrays 
2. Adding an additional external.configuration property to the config that's set like I did in the pull request earlier while keeping html_attributes as is

I can implement this and add a pull request for it but would like to check which way you prefer first? Or do you have another solution for it? Am I missing something here?

shalvah:
My guess is that it causes an error because the JSON includes double quotes, which terminate the attribute string. If we escape the quotes, we should be good. Try `addslashes`.

TS1997:
I tried that and still got the same error but did manage to solve it by doing the following:
```php
'external' => [
    'html_attributes' => [
[...]

PR #952, opened 2025-02-19
Updated scalar view to support configuration using html_attributes config

TS1997:
I was having issues configuring Scalar using the html_attributes property in the config which I created issue #949 for. After doing a bit of debugging I found out that the Scalar view didn't set the html_attributes property as the config properly and have updated the view to support this

### DIFF
--- a/resources/views/external/scalar.blade.php
+++ b/resources/views/external/scalar.blade.php
@@ -1,27 +1,24 @@
 <!doctype html>
 <html>
+
 <head>
     <title>{!! $metadata['title'] !!}</title>
-    <meta charset="utf-8"/>
-    <meta
-        name="viewport"
-        content="width=device-width, initial-scale=1"/>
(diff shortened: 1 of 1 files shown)


shalvah:
Oh, interesting. So scalar doesn't use direct attributes? This does seem to break the definition of "html_attributes", as we are no longer passing them directly to the element. Wouldn't it break things such as `data-proxy-url`?  Do you have a doc link for this Scalar behaviour?

I'm leaning towards leaving this as it is. After all, `$element.dataset.configuration` should be equivalent to a `data-configuration` attribute.

TS1997:
After looking into this a bit further it does look like a data-configuration attribute should work but isn’t recommended as it can get large but that shouldn’t be an issue since you’ve moved it to a config: https://guides.scalar.com/scalar/scalar-api-references/html#configuration

I must’ve completely missed that yesterday and will try that when I get home

Issue #343, opened 2021-10-20
respecting the sometimes rule in form request


khaled-mohamed2021:
- [x] **I confirm that I have read and attempted the tips in the [Troubleshooting Guide](https://scribe.knuckles.wtf/laravel/troubleshooting).**

**What happened?**

i am using the sometimes validation rule but the documentation not respecting it and always indicates that is required 

![image](https://user-images.githubusercontent.com/92052784/138103072-4dc1e20c-00f3-4451-96f4-726e371b6cc9.png)

![image](https://user-images.githubusercontent.com/92052784/138103169-9e7e49b8-6be1-4205-beec-ef7645a8f147.png)

khaled-mohamed2021:
@shalvah actually it did for me , i didn't even know about that rule , thank u for the new info , also keep up the *more than great* job . appreciate your efforts

PR #958, opened 2025-02-25
feat: add support to 'sometimes' rule

max13fr:
Hello,

I added the support for the "sometimes" validator rule that allow to skip all following validators like "required".

Currently, a field with this following rules was consired as required (but it shouldn't) : 
- `'consent_cgu' => 'sometimes|accepted'`
- `'other_field' => 'sometimes|required|...'` ("filled" rule seems better for this case but "filled_if", "filled_with", ... not existing so it's not an universal solution).

It notably fix this issue without any change to follow Laravel behaviour : https://github.com/knuckleswtf/scribe/issues/343

As Laravel behaviour, the following rules is correctly returned as required : `'badrule'=> 'required|sometimes|...'`

Best regard

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -79,8 +79,10 @@ trait ParsesValidationRules
 
                 // First pass: process rules which provide no type or example info
                 $firstPassRuleNames = [
+                    "sometimes",
                     "required",
                     "required_*",
+                    "accepted",
                     "same",
                     "different",
                     "nullable",
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -284,9 +284,9 @@ trait ParsesValidationRules
                     $parameterData['setter'] = fn() => true;
                     break;
 
-                    /*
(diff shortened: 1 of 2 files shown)


shalvah:
Hey, sorry I missed this. I'm a bit reluctant to add another parameter, `sometimes` to the validation logic. Why not simplify this to: if `sometimes` is provided, set `required` to false (unless it was already true) ?

max13fr:
Hi, no worries.

> Why not simplify this to: if sometimes is provided, set required to false (unless it was already true) ?

Because the case "sometimes|accepted" would not be managed for example : 
- first "sometimes" set required to false
- then "accepeted" set required to true

As "sometimes" is almost always at the beginning, changing required to false is useless (since it is already false by default) but on the other hand it risks being overwritten behind (accepted, required, ...).

shalvah:
Can  you add a test?

max13fr:
Hello,

Sorry for the delay, I've been very busy these past few months.

I just added the tests.

Best regards

shalvah:
> I've been very busy these past few months.

Haha, so have I! Thanks!

PR #972, opened 2025-03-21
Fix issue with route scope binding when getRouteKeyName method is different from Route scoped

Joemires:
There's a problem when you are using 
```php
Route::apiResource('products', Controllers\Store\ProductController::class)->scoped(['product' => 'slug'])
```
The documentation still uses your ```getRouteKeyName``` model method and this is a great problem for people with special routes having their special scopes


shalvah:
I probably only know about 30% of Laravel features at this point. 😅

shalvah:
Please fix the tests + add a new one for this use case, if possible.

shalvah:
Tests fixed (wasn't your fault), but can you add a test for this too?

Joemires:
Hey @shalvah 
Sorry I was busy with things

I'll look into this and revert

PR #908, opened 2024-10-28
Format OpenAPI response codes as strings

twickstrom:
This pull request fixes https://github.com/knuckleswtf/scribe/issues/907

By making sure the the responses status code keys are formatted as a `string`

It also adds `contact` to the info section of the `openapi.yaml` file.
It also correctly injects the `version` into the info section (previously it was static 1.0.0).

```
'info' => [
    'title' => $this->config->get('title') ?: config('app.name', ''),
    'description' => $this->config->get('description', ''),
    'version' => config('app.version', ''),
    'contact' => [
        'name' => $this->config->get('contact.name', ''),
        'url' => $this->config->get('contact.url', ''),
        'email' => $this->config->get('contact.email', ''),
    ],
],
```

Issue #907, opened 2024-10-28
OpenAPI Response YAML file response status codes most be a string


twickstrom:
### Scribe version

4.38

### PHP version

8.3.12

### Framework

Laravel

### Framework version

11.29

### Scribe config

```ruby
title => "The Fabric API"
base_url => "http://127.0.0.1:8000"
type => "external_static"
theme => "scalar"
static.output_path => "storage/scribe/docs"
laravel.add_routes => false
auth.enabled => true
auth.default => true
auth.name => "Authorization"
logo => "https://assets.whatsfabric.com/images/fabric-red-horizontal.svg"
strategies.queryParameters => added App\Docs\Strategies\QueryParameters\AddJsonApiQueryParameters
```

### What happened?

My `openapi.yaml` file is failing validation and is not able to be used by some third party tools for easier front end developement.

> OpenAPI can be represented canonically in either JSON or YAML, as you say in JSON only strings can be keys. With regard to YAML:
> 
> This field MUST be enclosed in quotation marks (for example, "200") for compatibility between JSON and YAML.
> 
> This has the effect that the key is always a string type.

See https://github.com/OAI/OpenAPI-Specification/issues/1752

The good news is this is a very easy fix.

`Knuckles\Scribe\Writing\Writer:27` 

Current:
`return Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP);`

New:
`return Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP | Yaml::DUMP_NUMERIC_KEY_AS_STRING);`

I would also recommend allowing OAS contact info to be added optionally:
`config\scribe.php`
```
'contact' => [
    'name' => 'Company Name',
    'url' => 'URL',
    'email' => 'EMAIL',
],
```
`Knuckles\Scribe\Writing\OpenAPISpecWriter`
Current:
```
'info' => [
    'title' => $this->config->get('title') ?: config('app.name', ''),
    'description' => $this->config->get('description', ''),
    'version' => '1.0.0',
],
```

New:
```
'info' => [
    'title' => $this->config->get('title') ?: config('app.name', ''),
    'description' => $this->config->get('description', ''),
    'version' => config('app.version', ''),
    'contact' => [
[...]

PR #979, opened 2025-04-08
Format OpenAPI response codes as strings

poldixd:
This is a smaller pull request derived from #908 .

More information can be found in issue/pull request #907 and #908.

### DIFF
--- a/src/Writing/Writer.php
+++ b/src/Writing/Writer.php
@@ -142,7 +142,7 @@ class Writer
         /** @var OpenAPISpecWriter $writer */
         $writer = app()->makeWith(OpenAPISpecWriter::class, ['config' => $this->config]);
         $spec = $writer->generateSpecContent($groupedEndpoints);
-        return Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP);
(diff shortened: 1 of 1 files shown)


src/Writing/Writer.php
```diff
@@ -142,7 +142,7 @@ public function generateOpenAPISpec(array $groupedEndpoints): string
         /** @var OpenAPISpecWriter $writer */
         $writer = app()->makeWith(OpenAPISpecWriter::class, ['config' => $this->config]);
         $spec = $writer->generateSpecContent($groupedEndpoints);
-        return Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP);
+        return Yaml::dump($spec, 20, 2, Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE | Yaml::DUMP_OBJECT_AS_MAP | Yaml::DUMP_NUMERIC_KEY_AS_STRING);
```
shalvah, line 145: Hmm, are we sure this doesn't break anything else? I'd much rather go the route of ensuring that the response status codes are actually strings, so they get dumped as strings without needing this.

shalvah:
Superseded by https://github.com/knuckleswtf/scribe/commit/80d21f1c4678e44ba8e2e549f075e7b3bfd72fe5

src/Writing/Writer.php
snuk182: Unfortunately, this change is still required. The current 5.8.0 version produces results, that do not pass the OpenAPI spec validation. Perhaps this can be formed as a config item to turn on or off.

PR #992, opened 2025-05-08
fix: Possible empty part of Route when path_param is not mandatory

### LINKED ISSUE #915: Optional Parameters not working when using prefixed group
### Scribe version

4.37

### PHP version

8.3.9

### Framework

Laravel

### Framework version

11.28

### Scribe config

```ruby
static.output_path => "docs"
laravel.add_routes => false
laravel.docs_url => "/"
auth.enabled => true
auth.default => true
auth.use_value => "xxxxx"
```

### What happened?

When generating the api docs, routes which

a) have an optional parameter
b) are prefixed by a Route::prefix()->group(...)

are missing prefixes and the optional parameter.

You can verify by creating this example:

```php
Route::prefix('users')->name('users.')->group(function () {
    Route::get('{user?}/show', fn () => 'hello')->name('show');
});
```

**Expected result:**
```
users/{user?}/show
```

**Actual result:**
```
//show
```

This issue does not appear however when using non optional parameters.

### Docs

s-shiryaev:
Hello

This PR fixes https://github.com/knuckleswtf/scribe/issues/915

When replacing with `str_replace()` , if `$search` and `$replace` have different number of elements, it is possible to replace with an empty string. 
This PR fixes it.

### DIFF
--- a/src/Extracting/Shared/UrlParamsNormalizer.php
+++ b/src/Extracting/Shared/UrlParamsNormalizer.php
@@ -62,7 +62,12 @@ class UrlParamsNormalizer
                 if (!$alreadyFoundResourceParam) {
                     // This is the first resource param (from the end).
                     // We set it to `params/{id}` (or whatever field it's bound to)
-                    $replaceWith = ["$pluralResource/{{$binding}}", "$pluralResource/{{$binding}?}"];
+                    $replaceWith = [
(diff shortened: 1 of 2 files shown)


shalvah:
Thanks for this! Can you add a test?

s-shiryaev:
Sure. Here it is! @shalvah 

Quite a rare case 🤔

PR #993, opened 2025-05-08
refactor: update the message on exist rule validation- form requests

alexlatam:
For security reasons it is important to prevent the database table names from being visible in the documentation. 

Because the documentation can be seen by external clients of the project.

### DIFF
--- a/src/Extracting/ParsesValidationRules.php
+++ b/src/Extracting/ParsesValidationRules.php
@@ -549,7 +549,7 @@ trait ParsesValidationRules
                     $parameterData['nullable'] = true;
                     break;
                 case 'exists':
-                    $parameterData['description'] .= " The <code>{$ruleArguments[1]}</code> of an existing record in the {$ruleArguments[0]} table.";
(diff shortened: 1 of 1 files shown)


src/Extracting/ParsesValidationRules.php
```diff
@@ -549,7 +549,7 @@ protected function processRule($rule, $ruleArguments, array &$parameterData, arr
                     $parameterData['nullable'] = true;
                     break;
                 case 'exists':
-                    $parameterData['description'] .= " The <code>{$ruleArguments[1]}</code> of an existing record in the {$ruleArguments[0]} table.";
+                    $parameterData['description'] .= " The <code>{$ruleArguments[1]}</code> value must exist.";
```
shalvah:
```suggestion
                    $parameterData['description'] .= " The <code>{$ruleArguments[1]}</code> of an existing record";
```

I'd rather go with this.

shalvah:
On second thought, I'm not sure I want to do this now.

PR #994, opened 2025-05-09
feat: Routes can be marked as deprecated

s-shiryaev:
In OpenAPI 3.0 it is possible to specify paths as deprecated. It is possible to implement this using custom spec generators, themes. 

But I think it would be useful to implement native support.

### Changes

1. Updated `Metadata` DTO
2. Updated Metadata strategies for DocBlock/Attributes
3. Updated base OpenApiSpecGenerator
4. Updated PostmanCollectionGenerator _(Since Postman doesn't have a way to mark methods as deprecated, I decided to add the [DEPRECATED] suffix to method names.)_
5. Updated `default` and `elements` themes
6. Updated tests

### Usage

Just add a annotation `@deprecated` in method/class/route DocBlock:
```PHP
/**
 * Update Role
 *
 * @deprecated
 */
public function update(string $id, CreateUpdateRoleRequest $request): Role
{
    $data = $request->validated();
    $model = $this->roleService->getModel($id, ['permissions']);

    return $this->roleService->updateRole($model, $data);
}
```
You can also use attributes:
```PHP
#[Deprecated]
public function create(CreateUpdateRoleRequest $request): Role
{
    $data = $request->validated();

    return $this->roleService->createRole($data);
}
 ```

### Examples

<details><summary>Default theme</summary></details>

<details><summary>Elements theme</summary>

</details>

<details><summary>External elements, scalar, rapidoc</summary>

</details>

### DIFF
--- a/camel/Extraction/Metadata.php
+++ b/camel/Extraction/Metadata.php
@@ -13,4 +13,5 @@ class Metadata extends BaseDTO
     public ?string $title;
     public ?string $description;
     public bool $authenticated = false;
+    public bool $deprecated = false;
 }
--- a/camel/Output/OutputEndpointData.php
+++ b/camel/Output/OutputEndpointData.php
@@ -313,6 +313,11 @@ class OutputEndpointData extends BaseDTO
         return $this->metadata->authenticated;
     }
 
+    public function isDeprecated(): bool
+    {
+        return $this->metadata->deprecated;
+    }
+
     public function hasJsonBody(): bool
     {
         if ($this->hasFiles() || empty($this->nestedBodyParameters))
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by oralunal** [comment]
I was doing this by using HTML helpers. This is a good feature.

```
   /**
     * Daily Insights (Deprecated)
     *
     * <small class="badge badge-grey">Deprecated</small>
     *
     * Retrieve daily insights. <i>Deprecation Note: Use GET method instead.</i>
     */
```

It would be great if we can give more information about the deprecation.

shalvah:
Wow, thank you for the thorough work.

PR #998, opened 2025-06-12
Made the bullet lists capitalization consistent.

BACON8ER-GAMER:
I edited the README so that the bullet lists would have consistent capitalization. Two of the three lists had each point start with capital letters, while one list did not. With this PR, the starting word of each bullet point is capitalized.


shalvah:
Thanks for paying attention the grammar! They are consistent though. Each of those is meant to form a complete sentence, together with the bullet point they're nested under, "Scribe can:"

PR #1001, opened 2025-06-27
Move intro_text directly after description in config/scribe.php for easier discovery

dazza-dev:
When you publish the default Scribe config, the intro_text option is located much further down the file—well after many unrelated settings. However, in the generated documentation this text appears immediately beneath the API description.
This separation makes the setting surprisingly hard to find for new users who naturally look for it near description.

Why it matters
Discoverability – Users tweaking the intro usually open the file, spot description, and expect the next few lines to let them customise the introduction itself.

Logical grouping – Both fields control the top-of-page content. Grouping them keeps the mental model consistent.

Developer ergonomics – Less scrolling/searching speeds up onboarding and day-to-day config edits.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -9,10 +9,18 @@ use function Knuckles\Scribe\Config\{removeStrategies, configureStrategy};
 
 return [
     // The HTML <title> for the generated documentation.
-    'title' => config('app.name').' API Documentation',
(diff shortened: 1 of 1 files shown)


shalvah:
I think it was there before, but I moved it because it was one of the larger items in the config file and I didn't want to overwhelm users out of the box. Your argument makes sense though. Thanks.

PR #1002, opened 2025-06-27
Add domain option to the laravel config section so docs can live on a separate hostname

dazza-dev:
Current behaviour
The laravel section of config/scribe.php lets you tweak docs_url, asset folder, middleware, etc., but there’s no way to put the generated documentation behind a different domain or sub-domain. Today you have to copy the route declaration and wrap it in Route::domain() yourself each time you publish the config.

Impact
Non-breaking: existing installations keep working as domain defaults to null.
Zero maintenance cost: publishing the config once exposes the new env var for developers who need it.

### DIFF
--- a/config/scribe.php
+++ b/config/scribe.php
@@ -61,7 +61,7 @@ return [
         'add_routes' => true,
 
         // Domain to use for the docs endpoint for example: https://docs.yourdomain.com/.
-        'domain' => null,
+        'domain' => env('SCRIBE_DOCS_DOMAIN', null),
 
         // URL path to use for the docs endpoint (if `add_routes` is true).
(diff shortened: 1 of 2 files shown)


shalvah:
Thanks for identifying a problem and trying to fix it! I won't be accepting this, though. 

As a rule, I reject any new additions to the config file. It's already quite big, and if an addition is not relevant to the majority of users, it's not worth increasing the size of this file (which hurts its usability).

Your use cases in particular can already be solved with a workaround: 

> Today you have to copy the route declaration and wrap it in Route::domain() yourself each time you publish the config.

I agree it may be less than ideal, so I'm open to other solutions (eg custom hooks). But the config file is for the 90% use case, not the 10%.

PR #1011, opened 2025-07-19
Add unique validation rule.

Ferks-FK:
This PR adds support for the `unique` rule.
Currently validations that use the `unique` rule were ignored.

With this PR:
```php
    public function rules(): array
    {
        return [
            'code' => ['required', 'string', 'alpha_dash', 'min:4', 'max:36',
                Rule::unique('vouchers')->ignore($this->route('voucher')?->id),
            ],
        ];
    }
```
image

Without (same rules):
image

Note: I tried to make `$this->route(‘voucher’)` resolvable, so that the description would be:
```
Must be unique in the <code>vouchers</code> table for column <code>id</code> (ignoring record with <code>id</code> = <code>5</code>).
```
This is useful for update cases where we need to ignore an ID.
However, I couldn't get it to work and I don't even know if it's possible, let me know.


src/Extracting/ParsesValidationRules.php
```diff
@@ -548,6 +548,24 @@ protected function processRule($rule, $ruleArguments, array &$parameterData, arr
                 case 'nullable':
                     $parameterData['nullable'] = true;
                     break;
+                case 'unique':
+                    $table = $ruleArguments[0] ?? 'table';
+                    $column = isset($ruleArguments[1])
+                        ? ($ruleArguments[1] === 'NULL' ? 'id' : $ruleArguments[1])
+                        : 'id'; // If the column is not specified, assume it is 'id'.
```
shalvah: I think doing this may cause some issues for people with some other column. Can we change this to not mention a column instead?

shalvah (commented):
Thanks for the work! You need to check the tests, though.

On a conceptual level, I'm not convinced about this feature-I've had some people complain about any references to a database table. 🤔 I think there's an argument there, that this is an implementation detail we shouldn't mention in the API. The more I think about it, the more I think it's better to just say "Must not be taken".

src/Extracting/ParsesValidationRules.php
Ferks-FK: As it stands, it does not require the column to be provided.
`Rule::unique('vouchers')` It is the equivalent `Rule::unique('vouchers', 'id')`
The ID column is assumed, because Laravel assumes that it is also ID.

shalvah: Let's just go with "Must not be taken" (which is Rails' default `unique` validation message). Database table and column names are implementation details that API consumers should neither know or care about.

shalvah (changes requested):
Let's go with "Must not be taken".

shalvah (changes requested):
Sorry, I still think we should not be mentioning the table name. We had something like that before, but there were complaints that we were exposing an implementation detail to API consumers. And that makes sense: if the endpoint is `POST /suppliers`, but you store in a table called `third_parties` (for instance, if you do Single Table Inheritance), returning an error message, `Must be unique in the "third_parties" table` would be confusing.

Sorry it's taken me so long to get back to you on this!

PR #1016, opened 2025-07-31
Add support for `list` validation rule

d3radicated:
This PR adds support for the `list` validation rule. Currently, the generated type for lists is `string`.

Just whipped this up on a whim. Willing to add tests but I don't know it it will break tests since `list` rule was only added on Laravel 11 and up.


shalvah:
For testing, just do something like this!
https://github.com/knuckleswtf/scribe/blob/66283b27b5a5f77877ad9c8660a5a68ba4a20729/tests/Unit/ValidationRuleParsingTest.php#L462-L468

PR #1019, opened 2025-08-21
Add deprecation attribute string support

### LINKED ISSUE #1018: Enhance the deprecated attribute
### Scribe version

5.3.0

### Your question

I really like the new attribute for the deprecation notice.
It's really helpful.
I think it could be even more helpful if we could add an optional deprecation date to add more context.

```php
#[Deprecated(deprecated: true, date: '2025-12-31')]
```
will render as
Image

If there is no date parameter, the behavior stays the same.

### Docs

- [x] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

Norda-AD:
Before, the deprecation badge was just a boolean. Now it can be used both as a boolean and a string. With a string, the badge is enhanced with said string to add some context. This does not work with external resources since the OpenAPI specification accepts only a boolean value.

Fix #1018

### DIFF
--- a/camel/Extraction/Metadata.php
+++ b/camel/Extraction/Metadata.php
@@ -13,5 +13,5 @@ class Metadata extends BaseDTO
     public ?string $title;
     public ?string $description;
     public bool $authenticated = false;
-    public bool $deprecated = false;
+    public bool|string $deprecated = false;
 }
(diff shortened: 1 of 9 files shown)


MielPoule:
+1 I am looking for this functionality as well !

shalvah:
Great work, thank you!

Issue #1032, opened 2025-10-22
Add nullable for ResponseFields in OpenAPI export


rene-marth:
### Scribe version

5.4.0

### PHP version

8.4

### Laravel version

 12.35.0

### Scribe config

```ruby
n/a
```

### What happened?

Currently the OpenAPI generation ignores the `nullable` parameter for ResponseFields. (which is valid for OpenAPI 3.0)

After digging the sources a naive "fix" would be:

1. Add `public $nullable;` to `camel/Extraction/ResponseField.php`
2. Under `generateSchemaForResponseValue()` in `src/Writing/OpenApiSpecGenerators/BaseGenerator.php` add

L568
```
$this->setDescription($schema, $endpoint, $path);

//either it's explicitly nullable or the value is null (don't know if this is legit?)
if ((isset($endpoint->responseFields[$path]->nullable) && $endpoint->responseFields[$path]->nullable) || $value === null) {
            $schema['nullable'] = true;
        }
```

However I also stumbled over you comment here https://scribe.knuckles.wtf/laravel/migrating#more-configurable-openapi-generation - would it be possible and do you recommend to use this approach to add the nullable parameter to responseField? Or do you prefer an enhancement like the above code?

### Docs

- [x] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

PR #1041, opened 2025-11-20
Openapi nullable response fields

rene-marth:
(Reworked) Fix for https://github.com/knuckleswtf/scribe/issues/1032

Note: Also suppresses empty `description` and empty `enum` when not given in ResponseField-Attribute

### DIFF
--- a/camel/Extraction/ResponseField.php
+++ b/camel/Extraction/ResponseField.php
@@ -26,4 +26,7 @@ class ResponseField extends BaseDTO
     public $example;
 
     public array $enumValues = [];
+
+    /** @var boolean */
+    public $nullable;
 }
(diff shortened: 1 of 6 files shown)


rene-marth:
```
-                                    'value' => 'text/html; charset=UTF-8'
+                                   'value' => 'text/html; charset=utf-8'
```

@shalvah I'm a little bit stuck at this failure - locally the tests run **only** if `UTF` (capital) is used. I'm using PHP8.4 here.

We could change the test to be case insensitive (`assertEqualsIgnoringCase`) but I'm unsure if this is valid. I'm still struggling to find the code that generates this header and why it generates `UTF-8` in my local system and `utf-8` here.

Could you help?

shalvah:
Hmm, I'm not sure. Where's the source of that string? What generates the content type?

rene-marth:
From what I found out this comes from the local php.ini. The content-type header itself is case insensitive, so it shouldn't matter if its upper- or lowercase. The test however checks case sensitive.

The latest commit tried to tackle this issue. My local tests completed successfully, however the test pipeline here in Github still fails...

shalvah:
Nice, thank you!

PR #1044, opened 2025-12-23
fix: Prevent empty enums when omitting ResponseField parameters

raphaelflash:
## Purpose
This PR fixes a bug where empty `enum: []` arrays were being added to the OpenAPI specification when using `ResponseField` attributes when only adding specific parameters (e.g. `required: true`).

## Problem
The default parameter value of `enum` is an empty array. In `OpenApiSpecGenerators/BaseGenerator`, the `generateSchemaForResponseValue()` method was using `isset()` to check for enum values. This means, that the check returns true, even if `enumValues` has not been explicitly set (empty array). This added empty enum arrays to the OpenAPI spec output.

## Impact
2 tests have been added to verify the following behavior:
- `ResponseField` without enum values
  - **Before**: `ResponseField` attributes with only `required: true` would generate OpenAPI specs with empty `enum: []` arrays
  - **After**: Empty enum arrays are no longer added to the OpenAPI spec
- `ResponseField` with actual enum values
  - No change

### DIFF
--- a/src/Writing/OpenApiSpecGenerators/BaseGenerator.php
+++ b/src/Writing/OpenApiSpecGenerators/BaseGenerator.php
@@ -567,7 +567,7 @@ class BaseGenerator extends OpenApiGenerator
         $this->setDescription($schema, $endpoint, $path);
 
         // Set enum values for the property if they exist
-        if (isset($endpoint->responseFields[$path]->enumValues)) {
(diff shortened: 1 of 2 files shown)


raphaelflash:
I've just seen that the same bug has recently been fixed within https://github.com/knuckleswtf/scribe/pull/1041. This means my PR now only adds new test cases.

shalvah:
Nice, thank you!

PR #1046, opened 2026-01-07
fix: Respect wrapping of API resources

raphaelflash:
## Purpose

This PR fixes the response field extraction to correctly respect Laravel's wrapping of API resources.

## Problem

I noticed that with nested API resources, the `ResponseField` attributes didn't correctly propagate their `required` status to the OpenAPI schema.

```php
class TestNestedOuterResource extends JsonResource
{
    #[
        ResponseField('outer1', required: true),
        ResponseField('outer1.inner1', required: true),
        ResponseField('outer2', required: true),
        ResponseField('outer2.inner2', required: true),
    ]
    public function toArray($request)
    {
        return [
            'outer1' => [
                'inner1' => 'string',
            ],
            'outer2' => new TestNestedInnerResource($this->resource),
        ];
    }
}
```

Setting the `inner` response fields to required did not mark them as required in the OpenAPI schema.

The reason was this fallback: https://github.com/knuckleswtf/scribe/blob/9b307b3cb617ee23f93e71dcef8f3da4c1c44060/src/Writing/OpenApiSpecGenerators/BaseGenerator.php#L607 Which always looked for the `$property` instead of the full path. Because the `data.` prefix was not considered when processing the ResponseField attributes.

## Changes

1. **`GetFromResponseFieldAttribute`**: Now respects `$wrap` property from API resources
2. **`GetFromResponseFieldTag`**: Now respects `$wrap` property when processing `@apiResource` tags
3. Both strategies now:
   - Read the `public static $wrap` property from the resource class
   - Apply the appropriate prefix to the field

Plus unit and integration test cases.

## Impact

With my changes
- Laravel's resource wrapping is respected
- It still works correctly when you customize the static $wrap property of the resource class
- It still works correctly when you disable wrapping via `JsonResource::withoutWrapping();`
- Response fields of nested resources are working correctly (see example with `TestNestedInnerResource`)

Hope this helps! 😊

### DIFF
--- a/src/Extracting/Strategies/ResponseFields/GetFromResponseFieldAttribute.php
+++ b/src/Extracting/Strategies/ResponseFields/GetFromResponseFieldAttribute.php
@@ -19,24 +19,51 @@ class GetFromResponseFieldAttribute extends PhpAttributeStrategy
 
     protected function extractFromAttributes(
         ExtractedEndpointData $endpointData,
-        array $attributesOnMethod, array $attributesOnFormRequest = [], array $attributesOnController = []
-    ): ?array
+        array $attributesOnMethod,
+        array $attributesOnFormRequest = [],
+        array $attributesOnController = []
+    ): ?array {
+        return [
+            ...$this->getNonApiResourceFields($endpointData, $attributesOnMethod, $attributesOnFormRequest, $attributesOnController),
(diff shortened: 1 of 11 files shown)


shalvah:
Thanks!

Issue #1058, opened 2026-02-18
Scribe 5.7 throws error exception Undefined array key "text"


troccoli:
### Scribe version

5.7.0

### PHP version

8.4.17

### Laravel version

12.49.0

### Scribe config

```ruby
title => "Hub API Documentation"
base_url => "https://hub.sourceglobalresearch.com"
theme => "elements"
try_it_out.enabled => false
auth.enabled => true
auth.default => true
auth.placeholder => "{TOKEN}"
postman.enabled => false
openapi.enabled => false
strategies.queryParameters => added App\Docs\Strategies\AddPaginationParameters
strategies.responses => added App\Docs\Strategies\UnauthorisedResponse
```

### What happened?

I have been using Scribe 5.6 for some time, but today after I upgraded to 5.7 I got the following error

```
app-php-webapp git:(hub-345-add-search-to-api) ✗ php artisan scribe:generate --verbose
  [GET] api/v1/sign-ins ............................................................................................................... 58.97ms DONE
  [GET] api/v1/searches ................................................................................................................ 7.79ms DONE
  [GET] api/v1/users ................................................................................................................... 7.89ms DONE
  [GET] api/v1/firms ................................................................................................................... 3.68ms DONE
  Extracting intro and auth Markdown files to: .scribe 

   WARN  Skipping modified file .scribe/intro.md.  

................................................................................. 4.86ms DONE
  Writing Blade docs to resources/views/scribe/ and assets to public/vendor/scribe/ .................................................... 1.23ms FAIL

In MarkdownParser.php line 16:
                              
  [ErrorException]            
  Undefined array key "text"  
                              

Exception trace:
  at /Users/giuliotroccoli-allard/Herd/app-php-webapp/vendor/knuckleswtf/scribe/src/Tools/MarkdownParser.php:16
 Illuminate\Foundation\Bootstrap\HandleExceptions->handleError() at /Users/giuliotroccoli-allard/Herd/app-php-webapp/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:258
 Illuminate\Foundation\Bootstrap\HandleExceptions->{closure:Illuminate\Foundation\Bootstrap\HandleExceptions::forwardsTo():257}() at /Users/giuliotroccoli-allard/Herd/app-php-webapp/vendor/knuckleswtf/scribe/src/Tools/MarkdownParser.php:16                                                                                                                                                                                                                                 
[...]

PR #1059, opened 2026-02-18
Fix undefined array key "text" in MarkdownParser when writing Blade docs

carlosrgzm:
## Problem
`scribe:generate` fails with `Undefined array key "text"` in `MarkdownParser::blockHeader` when writing Blade docs. Parsedown stores ATX header text in `element['handler']['argument']`, not `element['text']`, so the code that assumed `element['text']` triggers the error.

## Solution
Resolve header text from `element['text']` if set, otherwise from `element['handler']['argument']` (Parsedown’s structure), with a string fallback so slug/headings are always defined.

### DIFF
--- a/src/Tools/MarkdownParser.php
+++ b/src/Tools/MarkdownParser.php
@@ -12,11 +12,15 @@ class MarkdownParser extends \Parsedown
     {
         $block = parent::blockHeader($Line);
         if (isset($block['element']['name'])) {
+            $text = $block['element']['text']
+                ?? $block['element']['handler']['argument']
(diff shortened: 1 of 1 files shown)


corazzi:
Thanks, can confirm this fix works for me. 
I believe it's a parsedown 1.8 issue, 1.7.x worked fine but the structure was changed.

Hope this can get merged soon 🤞

sgrgrsn:
Fixes [this issue](https://github.com/knuckleswtf/scribe/issues/1058)

PR #1068, opened 2026-03-23
Fix: Move required from array schema to object schema inside items

gregmillz:
In the openapi spec, the required array belongs on the object schema inside items, not on the parent array schema ([see data types](https://swagger.io/docs/specification/v3_0/data-models/data-types/#objects))

- Fix `required` field placement in array item schemas. It was being set on the array schema itself instead of inside `items`, but it should have been in the `object` schema. The fix for this is in the function `generateSchemaForResponseValue`.

Example:
Incorrect: `required` object properties are in the array schema
```
lineItems:
  type: array
  required:  # `required` is not a valid keyword on type: array
    - productId
    - quantity
  items:
    type: object
    properties:
      productId:
        type: string
      quantity:
        type: integer
      note:
        type: string
```
Correct: `required` object properties are in the object schema:
```
lineItems:
  type: array
  items:
    type: object
    required:
      - productId
      - quantity
    properties:
      productId:
        type: string
      quantity:
        type: integer
      note:
        type: string
```

- Add missing `required` field support for responses that are arrays of objects. For example, when Laravel APIs return a collection, it's wrapped in a `data` key like `{ "data": [{"name": "foo"}]}` but you can call `JsonResource::withoutWrapping()` to get an array instead: `[{"name": "foo"}]`. The fix for this is in `generateResponseContentSpec`.

### DIFF
--- a/src/Writing/OpenApiSpecGenerators/BaseGenerator.php
+++ b/src/Writing/OpenApiSpecGenerators/BaseGenerator.php
@@ -485,20 +485,20 @@ class BaseGenerator extends OpenApiGenerator
 
                 // Non-empty array
                 if (is_object($decoded[0])) {
-                    // If the first item is an object, we assume it's an array of objects'
+                    // If the first item is an object, we assume it's an array of objects
                     $properties = collect($decoded[0])->mapWithKeys(function ($value, $key) use ($endpoint) {
                         return [$key => $this->generateSchemaForResponseValue($value, $endpoint, $key)];
                     })->toArray();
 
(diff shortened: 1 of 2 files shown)


gregmillz:
I see this is the same issue as [this PR](https://github.com/knuckleswtf/scribe/pull/1064), but I would like to see if the test changes are fixed with this pr

gregmillz:
CI is fixed by separate this other PR: https://github.com/knuckleswtf/scribe/pull/1071/changes

gregmillz:
Hi @shalvah! Is there a release schedule? No rush. The package is great, so would like to help improve it with this change :)

Issue #974, opened 2025-03-24
Private information doesnt expected at public docs


pheral:
### Scribe version

4.40.0

### Your question

Name of tables or models doesnt expected at public docs.

If we have request like this:
```php
/**
 * Query parameters.
 */
class SecureRequest extends FormRequest
{
  public function rules(): array
  {
    return [
      'request_key' => 'nullable|string|exists:'.SecureModel::class.',secure_column'
    ];
  }

  public function queryParameters(): array
  {
    return [
      'request_key' => ['description' => 'Your access key'],
    ];
  }
}
```

Scribe generate something like that:

> **request_key** string optional.
> Your access key. **The secure_column of an existing record in the App\Models\Secure table.**

but expected anoter result:

> **request_key** string optional.
> Your access key.
> 

And now we cant disable this secure information.

### Docs

- [x] I've checked [the docs](https://scribe.knuckles.wtf/laravel), the [troubleshooting guide](https://scribe.knuckles.wtf/laravel/troubleshooting), and [existing issues](https://github.com/knuckleswtf/scribe/issues?q=is%3Aissue+), but I didn't find a solution

shalvah:
Hmm. Scribe's behaviour is to _append_ to your description. We can make this configurable, but until then, you can remove the `QueryParameters\GetFromFormRequest::class` from the list of strategies in your config.

pheral:
All other appends from request rules - works great. 

That issue only about "exists" rule, because only that rule show to user secure information about database. 
Database - another layer, differs from request. 
And if in future you will make this rule configurable, all be fine.
We will just lock version 4.37.2 and wait for your updates.

shalvah:
Hmm, sure. In future, we'll make all validation rules strategies configurable. So you can specify: `"ignore_rules" => ["exists"]`.

andrewskm:
The exists rule is nice to show since it lets the user know it needs to, exist.
The issue is that the string for the message exposes the arguments that only the developer should be aware of and is risky to show publicly.

Would it be better to replace the string 
`The <code>{$ruleArguments[1]}</code> of an existing record in the {$ruleArguments[0]} table.`
To something more vague like
`The value of a column in an existing stored record.`

shalvah:
Yes, for sure! In fact, we can just replace it with something like `Must match an existing value`.

[...]

PR #1069, opened 2026-03-24
Update description for 'exists' validation rule

andrewskm:
Removed references to database attributes that should not be public.

Per https://github.com/knuckleswtf/scribe/issues/974, (only the private portion, not the ignoring feature).


shalvah:
I like it, although it will technically be a breaking change. 🤔

